PR #36, opened 2025-07-23
Class Wrapping for Nodes

### LINKED ISSUE #34: Extend Node Initialization Capabilities to Classes
Currently, Noob only supports wrapping functions with `WrapNode` classes.

Attempting the other two options cause implicit and explicit errors:

1. Passing class inheriting `Node`: signals are not gathered, and the outputs are getting lumped together and named `value`.

2. Passing a third-party class: `NotImplementedError`

1 is the main concern - this is due to the signal collection method sitting in `WrapNode` instead of `Node` class. The solution would be to move these (`collect_signal_names` and `_collect_signal_names`) to `Node`. This, however could force a somewhat ugly signal assignment as `Node`'s signal assignment either needs to happen in:
```python
def from_specification(cls, spec: "NodeSpecification") -> "Node":
    obj = resolve_python_identifier(spec.type_)
    params = spec.params if spec.params is not None else {}

    if inspect.isclass(obj):
        if issubclass(obj, Node):
            node = obj(id=spec.id, spec=spec, **params)
            node._signals = self.collect_signal_names(node.process)  # Added line. Not a fan
            return node
```
or,
```python
def model_post_init(self, __context: None = None) -> None:
    self._signals = self.collect_signal_names(self.process)  # Overridden in WrapNode
```

raymondwjang:
ref: #34 

## Summary

This PR enables using classes as a node, whether it's a subclass of the Node base class or not.

## Details

- Previous implementation missed the return signal name in the node attribute when using a `Node` subclass, ignoring the annotation and defaulting to `['value']`. This has been corrected / tested by moving `signal` attribute generation to `Node` class from `WrapNode`.

- Introduced an elementary implementation of `Node` coercion when the class is _**not**_ a `Node` subclass.


sneakers-the-rat (commented):
some minor changes discussed in person

sneakers-the-rat (approved):

PR #40, opened 2025-09-04
Event callbacks

raymondwjang:


tests/test_runner.py
```diff
@@ -0,0 +1,21 @@
...
+
+    tube = Tube.from_specification("testing-basic")
+    runner = SynchronousRunner(tube)
+
+    runner.init()
+    events = []
+
+    def _cb(event):
+        if event["node_id"] == "whatever":
+            return
```
sneakers-the-rat, line 14:
```suggestion
```

pyproject.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
     "pydantic-settings>=2.8.1",
     "platformdirs>=4.3.7",
     "PyYAML>=6.0.2",
+    "anyio>=4.10.0",
```
sneakers-the-rat, line 15: doesn't look like we are using this yet. maybe hold off adding it until we do, but fine leaving it in bc we will eventually anyway

sneakers-the-rat (approved):
lint this and good to go

Issue #43, opened 2025-10-15
Event-driven topological sorter


sneakers-the-rat:
Depends on: #21 

Currently we just rely on python's builtin `graphlib.TopologicalSorter` to determine whether a node should be run: https://github.com/miniscope/noob/blob/c96db36ebed8865606ab6b4ee77e8c1f2d0554f7/src/noob/tube.py#L90

the order of operations is:
- create graph
- get ready nodes
- run nodes
- get events
- store events in event store
- use events to update state of topo graph
- goto 0

but this doesn't work for a multiprocessing runner: `process` epochs are no longer completed before starting another processing epoch. so with the current impl., we would start another run through the graph, but then the events still left from the prior epoch would just be "orphaned" and never complete.

So we need to write our own topo sorter that
- receives events to update the state of the graph
- yields sets of node ids that are ready *and* an "epoch" that they are run in.

an "epoch" is one set of node processing operations that correspond to one "source" node output. or, generally, to the most recent cardinality limiting step. The epoch is used to resolve the problem of knowing which events "match" one another or should be aligned across multiple event emissions from the source (e.g. say we have a camera, and it is split into frame data and metadata, we would always want to process the metadata for a matching frame in some downstream `merge` node)

It is currently only well-defined for tubes with 
- one source node (node with no dependencies)
- multiple source nodes that are synchronized (yield the same number of events per unit time)

For example:
- Linear graph with a single source is simple: each time the source node is called the epoch is incremented, and all events emitted downstream of the sink just inherit the epoch
- In the case of a mapping operation that expands cardinality, where for e.g. 1 event is emitted from `source_a` and then that is mapped to 5 events from `node_b`, the epoch for each of those events would need to be hierarchical beneath the source node to track "matching" events (epoch `1.1`, `1.2`, etc.). If there is then a `gather` node that collects all the mapped outputs from epoch 1, the event it emits would "collapse" back down to just be `1.1`, etc.

epoch is orthogonal to event ID, which should just be an opaque unique identifier to refer to/select/etc. an event.

We want to be careful about letting state splay across a bunch of different objects, but the `Tube` is the most logical thing to keep track of the tube state. So the `Tube` should create a `TopologicalSorter` and use the information from emitted events to determine which nodes are ready to run. Since the `Tube` owns the instantiated `Node` objects, which can be stateful, the `Tube` also owns the `TopologicalSorter` graph thing (which we should probably call `Scheduler` because that's what it's doing now).

[...]

PR #46, opened 2025-10-22
Topological Sorter with Epoch

raymondwjang:
Ref: #43 

Building a separate `Scheduler` class for an event-driven topological sorter, that can handle epoch count (different set of "iterations"). It functions as a clock for epoch, which can be subsequently consumed by `EventStore`.

The initial implementation is a dictionary of `[epoch, graphlib.TopologicalSorter]`. Epoch is managed by `count` from `itertools`.

### DIFF
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -30,9 +30,15 @@ def test_disabled_node() -> None:
 
     assert set(runner.tube.enabled_nodes.keys()) == {"a", "c"}
 
-    # Return node is hanging and tries to index None
-    with pytest.raises(IndexError):
-        runner.run()
+    all_events = []
+
+    def _cb(event) -> None:
+        nonlocal all_events
+        all_events.append(event)
+
+    runner.add_callback(_cb)
+    runner.run()
+    assert "b" not in {e["node_id"] for e in all_events}
 
 
 def test_disabled_return() -> None:
@@ -56,13 +62,12 @@ def test_dynamic_disable_node() -> None:
     assert first == list(range(0, iters * 2, 2))
 
     runner.disable_node("b")
-    with pytest.raises(IndexError):
-        runner.run(n=iters)
+    assert runner.run(n=iters) is None
 
     runner.enable_node("b")
     result = runner.run(n=iters)
-    # continues from where it left off after the first run + one failure with disabled 'b' run.
-    start = max(first) + 4
+    # the tube has been running, just not returning anything
+    start = 400
     expected = list(range(start, start + (iters * 2), 2))
     assert result == expected
 
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -26,6 +26,21 @@ def test_epoch_increment():
         assert isinstance(scheduler[i], TopologicalSorter)
 
 
+def test_tube_increments_epoch(no_input_tubes):
+    """
+    Multiple runs of a tube should increment the epoch
+    """
+    tube = Tube.from_specification(no_input_tubes)
+    runner = SynchronousRunner(tube)
+
+    for i in range(5):
+        _ = runner.process()
+        events = runner.store.events
+        # we haven't cleared events
+        assert len(events) > 0
+        assert all(e["epoch"] == i for e in events)
+
+
 def test_event_store_filter():
     """
     Must be able to filter events in EventStore by epoch.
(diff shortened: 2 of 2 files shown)


src/noob/scheduler.py
```diff
@@ -0,0 +1,130 @@
...
+        are marked "done."
+
+        Not implemented in :class:`.Scheduler` while we're still utilizing
+        :class:`.TopologicalSorter`, since its API access to successors /
+        predecessors is limited.
+
+        """
+        raise NotImplementedError()
+
+    def add_graph(self, nodes: dict[str, Node], edges: list[Edge]) -> None:
```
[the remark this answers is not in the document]
sneakers-the-rat: ah, reading more, i misunderstood what `_graphs` was. disregard. i think we can do this later if we want to

sneakers-the-rat: it's a *little* awkward to me to need to pass the nodes each time, i would expect `add_graph` to mean "i add a set of nodes and edges, and now that i know about the structure of the graph, i can make `TopologicalSorters` from that. i get why this would be good to not make the scheduler need to have `disable/enable` and other methods, but it seems like it will result in a lot of repeated operations.

src/noob/scheduler.py
```diff
@@ -0,0 +1,130 @@
...
+            the dependent nodes. Disable nodes at your own risk.
+
+            This behavior will likely change,
+            allowing explicit parameterization of how optional values are handled,
+            see: https://github.com/miniscope/noob/issues/26,
+
+        """
+        sorter = TopologicalSorter()
+        enabled_nodes = [node_id for node_id, node in nodes.items()]
+        for node_id in enabled_nodes:
```
sneakers-the-rat: did we not have filtering for enabled nodes here?

src/noob/store.py
```diff
@@ -66,10 +70,14 @@ def get(self, node_id: str, signal: str) -> Event | None:
 
         ``None`` in the case that the event has not been emitted
         """
-        event = [e for e in self.events if e["node_id"] == node_id and e["signal"] == signal]
+        event = [
+            e
+            for e in self.events
```
sneakers-the-rat: i can feel this iteration getting more expensive... i think eventually this will need to be turned into a dictionary so we don't need to iterate through every single event every time

tests/test_scheduler.py
```diff
@@ -0,0 +1,36 @@
...
+    Scheduler must be able to hold multiple graphs of different statuses
+
+    """
+    scheduler = Scheduler()
+    noob_ids = ["testing-basic", "testing-basic", "testing-branch", "testing-disable-node"]
+    for id_ in noob_ids:
+        tube = Tube.from_specification(id_)
+        scheduler.add_graph(nodes=tube.enabled_nodes, edges=tube.edges)
+
+    assert len(scheduler.graphs) == len(noob_ids)
```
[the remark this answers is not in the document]
sneakers-the-rat: this doesn't make a ton of sense to me as a test without confirming stuff about how they work, like if we want to support this we should also test that we correctly yield generations of nodes from each added graph. i am not sure how we would differentiate them as being from different graphs - tubes do have `id` fields, so we would naturally extend our notion of signal/slot identifiers from `{node_id}.{signal_id}` to `{tube_id}.{node_id}.{signal_id}` and similar in `Event` objects? for now i would just say we shouldn't support this until we do the whole thing and have some design on how running multiple tubes at once might work.

sneakers-the-rat (commented):
ok ya a couple changes i think in how the scheduler needs to handle epochs, i can work on this tomorrow with you. 

and yes re: your prior comment, so when the TopologicalSorter for a given epoch completes, we should return some `EpochFinished` event in the list of (node, epoch) ready identifiers, and the scheduler should probably delete that topological sorter to avoid an infinitely growing dict

tests/test_scheduler.py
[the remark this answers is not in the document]
sneakers-the-rat: to do [this](https://github.com/miniscope/noob/pull/46#discussion_r2458571204), we might want to do this:

```python
class Graphs(TypedDict):
    nodes: list[Node]
    edges: list[Edges]
    source_nodes: list[NodeID]
    epochs: dict[int, TopologicalSorter]
```

src/noob/runner.py
```diff
@@ -229,32 +227,33 @@ def process(self) -> ReturnNodeType:
         """
         self.store.clear()
 
-        graph = self.tube.graph()
-        graph.prepare()
+        scheduler = self.tube.scheduler()
```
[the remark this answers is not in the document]
raymondwjang: makes sense

src/noob/scheduler.py
raymondwjang: Yeah i think the former makes sense since we would also be dynamically switching that as well.

src/noob/store.py
raymondwjang: like a dict with the keys holding the event information?

tests/test_scheduler.py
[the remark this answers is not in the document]
raymondwjang: yeah no this test is from me not having a correct understanding of how an asyncrunner would work 💀 will fix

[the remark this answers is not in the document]
sneakers-the-rat: Yeah if I'm getting your meaning of independent nodes as nodes without dependencies, then independent nodes == source nodes

src/noob/store.py
sneakers-the-rat: Oh yep, obvi that is the way

sneakers-the-rat: Wait that blogpost is needlessly complex. Just do

```python
{("node-id", "signal", 0): Event()}
```

sneakers-the-rat: Wait no thats the same problem BC wed need to iterate through all the keys, idk we can do whatever, and its probably not even that slow

tests/test_runner.py
```diff
@@ -30,9 +30,15 @@ def test_disabled_node() -> None:
...
-        runner.run()
+    all_events = []
+
+    def _cb(event) -> None:
+        nonlocal all_events
+        all_events.append(event)
+
+    runner.add_callback(_cb)
+    runner.run()
+    assert "b" not in {e["node_id"] for e in all_events}
```
sneakers-the-rat: directly testing that b was not run

src/noob/node/return_.py
```diff
@@ -35,7 +35,7 @@ def get(self, keep: bool) -> Any | None:
         """
         try:
             # FIXME: what a nightmare - make all of these derive from the spec
-            if isinstance(self.spec.depends, str):
+            if self._args and isinstance(self.spec.depends, str):
```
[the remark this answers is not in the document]
raymondwjang: Sounds good

PR #54, opened 2025-10-28
Zmq runner

sneakers-the-rat:
Behold: a p2p multiprocessing runner.

So we can't really just use a process pool or concurent futures or whatever because we 
a) don't want to and 
b) can't 
pickle every node every time we want to call it. 

we also want to eventually be able to run things on different machines.

so here's that: controlled p2p node runner.

- one command node is basically the "tracker" that starts the other nodes and connects them to each other
- then each node runs in its own process, subscribes to the other nodes that it consumed events, from, runs, then poops out more events.

it's pretty dang simple. 

so this is still a WIP, listing todos:
- [x] outer `ZMQRunner` class that creates the `CommandNode` and spawns the `NodeRunner`s
- [x] handling `return` nodes: subscribe to the returned events and yield as a generator
- [x] main loop inside of `NodeRunner` that just pulls from the event store as event combinations are ready for a given epoch
- [x] teardown
- [x] ser/des for non-jsonable events
- [ ] testing baby
- [x] Allow `EventStore.add()` to take an already-created event, rather than creating the event. 

caveats:
- uses IPC so won't work on windows
- intended for multiprocessing on one machine for now, will do multi-machine in another stage.

### DIFF
--- a/src/noob/scheduler.py
+++ b/src/noob/scheduler.py
@@ -12,6 +12,7 @@ from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
 
 from noob.const import META_SIGNAL
 from noob.event import Event, MetaEvent, MetaEventType, MetaSignal
+from noob.exceptions import EpochCompletedError, EpochExistsError
 from noob.logging import init_logger
 from noob.node import Edge, NodeSpecification
 from noob.types import NodeID
@@ -71,9 +72,9 @@ class Scheduler(BaseModel):
                 this_epoch = next(self._clock)
 
             if this_epoch in self._epochs:
-                raise ValueError(f"Epoch {this_epoch} is already scheduled")
+                raise EpochExistsError(f"Epoch {this_epoch} is already scheduled")
             elif this_epoch in self._epoch_log:
-                raise ValueError(f"Epoch {this_epoch} has already been completed!")
+                raise EpochCompletedError(f"Epoch {this_epoch} has already been completed!")
 
             graph = self._init_graph(nodes=self.nodes, edges=self.edges)
             graph.prepare()
--- a/src/noob/exceptions.py
+++ b/src/noob/exceptions.py
@@ -16,6 +16,10 @@ class ConfigError(NoobError):
     """Base config error type"""
 
 
+class SchedulerError(NoobError):
+    """Base error in the scheduler"""
+
+
 class StoreError(NoobError):
     """Base store error type"""
 
@@ -69,4 +73,16 @@ class ExtraInputWarning(InputWarning, RuntimeWarning):
     """
     Extra input was provided in some scope where it was not specified,
     but it was ignorable
+    """
+
+
+class EpochCompletedError(SchedulerError, ValueError):
+    """
+    An epoch was already completed, but some attempt was made to update it or use it.
+    """
+
+
+class EpochExistsError(SchedulerError, ValueError):
+    """
+    Epoch already exists and is active, but attempted to create it.
     """


src/noob/scheduler.py
sneakers-the-rat: just note to self, need to undo a lot here. i ultimately don't know if we want to have two 'modes' but if we do the difference between them will need to be a whole lot clearer. for now it's just a grab bag of hacks

sneakers-the-rat: removed the scheduler mode and just handled what needed to be done in the runner

src/noob/scheduler.py
```diff
@@ -188,10 +378,7 @@ def _init_graph(nodes: dict[str, NodeSpecification], edges: list[Edge]) -> Topol
             required_edges = [
                 e.source_node
                 for e in edges
-                if e.target_node == node_id
-                and e.required
-                and e.source_node in enabled_nodes
```
sneakers-the-rat: removed source node check here because a target node should especially not run if its source node is disabled - i.e. we need to keep its dependency on the source node

sneakers-the-rat:
ok marking as ready for review, did some cleanup so there's not junk/double implementations. still getting sporadic test failures on some of the runners so i think i'll need to do #73 so that we can debug that because it's running locally. pretty slow (lots of init/teardown time, but the actual runner is reasonable once it's running, the linux tests still run in liks 5 seconds lol). but pieces in place

sneakers-the-rat:
just added the ability to report status, did #73 , and now things are working pretty nicely. i think last thing i need to do here is make mypy happy

src/noob/types.py
```diff
@@ -105,7 +141,21 @@ def _not_reserved(val: str) -> str:
 """
 Union of all types of config sources
 """
-NodeID: TypeAlias = Annotated[str, AfterValidator(_is_identifier)]
+
+SerializableDatetime = Annotated[
+    datetime, BeforeValidator(_from_isoformat), PlainSerializer(_to_isoformat, when_used="json")
+]
+TPickle = TypeVar("TPickle")
```
raymondwjang: can `NewType` not do this?

raymondwjang: (just out of curiosity)

src/noob/runner/base.py
```diff
@@ -28,10 +30,24 @@ class TubeRunner(ABC):
...
+    """The max number of times that `iter` will call `process` to try and get a result"""
 
     _callbacks: list[Callable[[Event | MetaEvent], None]] = field(default_factory=list)
 
-    _logger: Logger = field(default_factory=lambda: init_logger("tube.runner"))
+    _logger: Logger = None  # type: ignore[assignment]
+    _runner_id: str | None = None
+
+    def __post_init__(self):
+        self._logger = init_logger(f"noob.runner.{self.runner_id}")
```
raymondwjang: why is this in `__post_init__` instead of `__init__`?

src/noob/types.py
```diff
@@ -74,6 +87,29 @@ def _not_reserved(val: str) -> str:
...
+
+def _to_isoformat(val: datetime) -> str:
+    return val.isoformat()
+
+
+def _to_jsonable_pickle(val: Any, handler: SerializerFunctionWrapHandler) -> Any:
+    try:
+        return handler(val)
+    except (TypeError, PydanticSerializationError):
+        return "pck__" + base64.b64encode(pickle.dumps(val)).decode("utf-8")
```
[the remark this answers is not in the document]
sneakers-the-rat: Gotta encode it somehow

src/noob/types.py
sneakers-the-rat: The typevar here is just a standin for whatever you do `Picklable[Type]`

src/noob/runner/base.py
sneakers-the-rat: BC its a data class, and overriding the init makes it not work no more, and we need the value of the runner id which is only defined after init

src/noob/runner/base.py
```diff
@@ -28,10 +30,24 @@ class TubeRunner(ABC):
...
+
+    @property
+    def runner_id(self) -> str:
+        if self._runner_id is None:
+            hasher = hashlib.blake2b(digest_size=4)
+            hasher.update(str(datetime.now(UTC).timestamp()).encode("utf-8"))
+            self._runner_id = f"{hasher.hexdigest()}.{self.tube.tube_id}"
```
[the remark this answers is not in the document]
sneakers-the-rat: Idk the datetime thing is probably not the best source for an ID, I just needed something to differentiate

src/noob/scheduler.py
```diff
@@ -100,49 +155,201 @@ def sources_finished(self, epoch: int | None = None) -> bool:
...
-                id=uuid4().int,
-                timestamp=datetime.now(UTC),
-                node_id="meta",
-                signal=MetaEventType.EpochEnded,
-                epoch=epoch,
-                value=epoch,
-            )
+        with self._ready_condition, self._epoch_condition:
+            if epoch in self._epoch_log:
+                self.logger.debug(
```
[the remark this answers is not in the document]
sneakers-the-rat: there shouldn't be. when we implement map, we'll need to make "sub-epochs" or something and figure out how we want to change the topo sorter (we definitely need to make our own by then) so that it wouldn't just be modeled as a single node in a single epoch. in the graph of work to do, something should only be marked done once in all cases i think.

src/noob/scheduler.py
```diff
@@ -100,49 +155,201 @@ def sources_finished(self, epoch: int | None = None) -> bool:
...
+        Returns:
+            int: the epoch that was completed.
+        """
+        with self._epoch_condition:
+            # check if we have already completed this epoch
+            if isinstance(epoch, int) and self.epoch_completed(epoch):
+                return epoch
+
+            if epoch is None:
+                self._epoch_condition.wait()
```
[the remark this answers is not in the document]
sneakers-the-rat: not sure what you mean - if epoch is `None` (we're waiting for *any* epoch to be completed) then we just wait until any epoch to be completed, which notifies the epoch condition, and then we just return the last value of the epoch log, which is the last epoch completed.

src/noob/runner/zmq.py
```diff
@@ -0,0 +1,725 @@
...
+        sub = self.context.socket(zmq.SUB)
+        sub.setsockopt_string(zmq.IDENTITY, "command.inbox")
+        sub.setsockopt_string(zmq.SUBSCRIBE, "")
+        sub = ZMQStream(sub, self.loop)
+        sub.on_recv(self.on_inbox)
+        return sub
+
+    def announce(self) -> None:
+        msg = AnnounceMsg(
+            node_id="command", value=AnnounceValue(inbox=self.router_address, nodes=self._nodes)
```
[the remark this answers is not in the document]
sneakers-the-rat: yes, i should have a docstring here - this is done after any node tells us about itself, where we store its information and then tell all the other nodes about it. then the node runners receive the announce and subscribe to the node's outbox if they depend on it. this allows us to only have one thing to configure, the command node, and then all the node runners are self-configuring.

src/noob/store.py
```diff
@@ -88,22 +140,48 @@ def collect(self, edges: list[Edge], epoch: int) -> dict | None:
...
 
         If some of the requested events are missing but others are present,
-        return ``None`` for any missing events.
+        exclude the keys from the returned dict
+        (`None` is a valid value for an event, so if a key is present and the value is `None`,
+        the event was emitted with a value of `None`)
+
+        If `epoch` is -1,
+        get the the events from the most recent epoch where all events are present,
+        and if no epochs are present with a full set of events,
```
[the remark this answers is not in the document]
sneakers-the-rat: https://github.com/miniscope/noob/pull/54/commits/445e3016d51341df3c2e2baf476a0223acf7fe4f

PR #64, opened 2025-11-21
temporary fix for gather infinitely populating epochs

raymondwjang:
Gather node handling has been jammed into syncrunner. maybe there's a better way...

### DIFF
--- a/src/noob/node/gather.py
+++ b/src/noob/node/gather.py
@@ -44,7 +44,7 @@ class Gather(Node, Generic[_TInput]):
     _items: list[_TInput] = PrivateAttr(default_factory=list)
     _lock: LockType = PrivateAttr(default_factory=Lock)
 
-    def process(self, value: _TInput, trigger: Any | None = None) -> list[_TInput] | None:
+    def process(self, value: _TInput, trigger: Any | None = None) -> list[_TInput] | NoEvent:
         """Collect value in a list, emit if `n` is met or `trigger` is present"""
         if trigger is not None and self.n is not None:
             raise ValueError("Cannot use trigger mode while `n` is set")
--- a/src/noob/scheduler.py
+++ b/src/noob/scheduler.py
@@ -137,7 +137,7 @@ class Scheduler(BaseModel):
             )
         return None
 
-    def end_epoch(self, epoch: int = None) -> None:
+    def end_epoch(self, epoch: int | None = None) -> None:
         if epoch is None:
             del self._epochs[list(self._epochs)[-1]]
         else:


src/noob/runner/sync.py
```diff
@@ -105,6 +106,11 @@ def process(self, **kwargs: Any) -> ReturnNodeType:
                 events = self.store.add(node.signals, value, node_id, epoch)
                 events = scheduler.update(events)
                 self._call_callbacks(events)
+
+                # not a fan of this...
+                if isinstance(node, Gather) and isinstance(events[0]["value"], NoEvent):
+                    scheduler._end_epoch(events[0]["epoch"])
```
sneakers-the-rat: wait why would this end the epoch?

raymondwjang: because while Gather node is gathering, it spawns a new epoch for each Gather. If we don't end the epoch when we hit Gather, we end up with a bunch of epochs that never finish

raymondwjang: or at least that's how the gather node works for now. i remember us talking about this when we were first implementing it 😅

sneakers-the-rat: hmm hmm. so i see what you're saying.

so say we have a graph like this

```mermaid
flowchart LR
  A --> Gather
  A --> B
  B --> C
  Gather --> D
```

We would not want to end the epoch after a non-emitting gather in all cases, because e.g. here we would yield sets of nodes like
- A
- B & Gather
- C & (optionally) D

and if we quit on a non-emitting gather then C would not get run despite that not having anything to do with the gather.

so the thing we want is for an epoch to run until there are no ready nodes remaining and then end the epoch. in the sync case we don't need to do anything special because we know the state of all the nodes.

raymondwjang: I think when Gather was emitting `None` the `Gather` node was perpetually ready since it was never being marked `.done`, and now that it's emitting `NoEvent`, node `D` is perpetually ready. I can check the details, but this was slowing down `Cala` since it was going though all `NoEvent` epochs it accrued and calling `.get_ready()` on them which was becoming hundreds of thousands..

raymondwjang: Makes sense. I just wasn't sure how to distinguish between "no more nodes are ready _yet_" vs. "no more nodes will _ever_ be ready anymore"...which i guess may not really be an issue we need to worry about in synchronous runner

raymondwjang: ok i think that solves the qualm i had about that line. i'll push a modification in a sec

raymondwjang: mmm would we want `NoEvent` events in the `EventStore`? My intuition says no (feels weird to say "no event" is an event) but are there any cases we'd want to hang a callback to it, or use it in some way in the zmq runner?

raymondwjang: in which case `NoEvent` should rather be a `MetaEvent`

src/noob/event.py
```diff
@@ -46,3 +46,12 @@ class MetaEvent(Event):
 
     node_id: Literal["meta"]
     signal: MetaEventType
+
+
+class NoEvent:
```
sneakers-the-rat: there's got to be a better way to do a non-numeric const in python, serializing this will mean some custom serialization i think

tests/test_runner.py
```diff
@@ -12,16 +11,23 @@ def test_process_callback() -> None:
...
 
-    assert len(events) == 2
+    # internal events are not caught in callback (or EventStore)
+    # Return node emits NoEvent
+    # NoEvent and MetaEvent are uncaught
+    assert (
+        len(cb_events)
+        == len(runner.store.events)
+        == len([node for node in tube.nodes.values() if not isinstance(node, Return)])
+    )
```
sneakers-the-rat:
```suggestion
    # internal events are not caught in callback (or EventStore)
    # Return node emits NoEvent
    # NoEvent and MetaEvent are uncaught
    assert (
        len(cb_events)
        == len(runner.store.events)
        == len([node for node in tube.nodes.values() if not isinstance(node, Return)])
    )
    assert len(cb_events) > 0
```

sneakers-the-rat (approved):
minor suggestions but would be fine as is.

the error in the tests is fixable by: https://github.com/miniscope/noob/pull/65/commits/fbb71ac19a4ad82b4d3cafb1144f401372da835b

PR #72, opened 2025-12-05
no event verse 2

raymondwjang:
singing the same verse

A few options in terms of how to implement NoEvent...

A. NOEVENT = object(): widely used. type annotation nightmare. tried NewType but NewType started bullying Pydantic when it became Union with other types, and nobody got a time for that.
B. a bare NoEvent class: feels weird. might accidentally instantiate it. also a hurdle in type annotation.
C. Singleton NoEvent class:

    Implementing a separate Singleton metaclass
    a NoEvent base class
    ???
    Profit

Not sure how this plays out with serialization still.

https://stackoverflow.com/questions/6760685/what-is-the-best-way-of-implementing-a-singleton-in-python

### DIFF
--- a/src/noob/event.py
+++ b/src/noob/event.py
@@ -8,6 +8,8 @@ if sys.version_info < (3, 12):
 else:
     from typing import TypedDict
 
+from noob.types import Singleton
+
 
 class Event(TypedDict):
     """
@@ -49,7 +51,7 @@ class MetaEvent(Event):
     signal: MetaEventType  # type: ignore
 
 
-class NoEvent:
+class NoEvent(metaclass=Singleton):
     """
     Signifier for no event emitted
 
--- a/src/noob/event.py
+++ b/src/noob/event.py
@@ -8,8 +8,6 @@ if sys.version_info < (3, 12):
 else:
     from typing import TypedDict
 
-from noob.types import Singleton
-
 
 class Event(TypedDict):
     """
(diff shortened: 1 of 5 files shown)


raymondwjang:
type annotation, actually... i could not figure out how to type annotate with the
```python
noevent = object()
```
format, because
```python
def _return() -> object:  # i hate returning type "object"
    ...
    return noevent
```

I tried `typing.NewType` but it was giving troubles with `Signal` type enforcement.

sneakers-the-rat:
i think that might be best, see how i'm handling this here: https://github.com/miniscope/noob/blob/a19beb58bb98bd5d6c5002bd01d03e48a028c215/src/noob/network/message.py#L97

(actually not exactly that commit, one second i just made some updates sot hat it wouldn't just always pickle events, but the idea is the same)

so the noevent message is like

```python
msg = message.EventMsg(
    type_="event", 
    node_id="example", 
    value=[
        Event(
            id=0, 
            timestamp=datetime.now(UTC), 
            node_id="example", 
            signal=META_SIGNAL, 
            epoch=0, 
            value=NoEvent()
        )
    ]
)

>>> as_json = msg.model_dump_json()
>>> pprint(json.loads(as_json))
{'node_id': 'example',
 'timestamp': '2025-12-05T02:47:49.273861Z',
 'type': 'event',
 'value': [{'epoch': 0,
            'id': 0,
            'node_id': 'example',
            'signal': '__META__',
            'timestamp': '2025-12-05T02:47:49.273561+00:00',
            'value': 'NOEVENT'}]}

```

so the enum could be `MetaSignals` or something

raymondwjang:
@sneakers-the-rat done!

sneakers-the-rat (approved):

Issue #49, opened 2025-10-24
Bubbling Up Errors


raymondwjang:
Noob has instances where Exceptions do not clearly point to which node / line it is coming from when `depends` / `params` entries in YAML spec are wrong.

I think it's dependent on how the nodes are wrapped, but it needs further investigation for now.

More details incoming.

sneakers-the-rat:
Ah do we need to do more validation when instantiating to check if the tube spec is correct? Can you give me a MWE of the problem?

sneakers-the-rat:
huh that shouldn't happen, a) we should validate the spec, and b) node `b` should never run because its dependencies should never be satisfied. [this line](https://github.com/miniscope/noob/blob/b940c0977711ad56e37e4aef7afbc9f374de374b/src/noob/scheduler.py#L193) is the problematic one there. 

if we do (a) then we don't need to worry about (b), and we should do (a).

PR #76, opened 2025-12-10
preserve original traceback in a note

sneakers-the-rat:
fix: #49

previously the error re-raised by the zmqrunner didn't have the traceback from the original error attached. it turns out it's a pain in the ass to create a synthesized traceback just for the purposes of printing, and we didn't want to add another dependency on tblib, so we just added it in a note. it's not *totally* clear, but it's something

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 1
----> 1 runner.process()

File ~/git/noob/src/noob/runner/zmq.py:670, in ZMQRunner.process(self, **kwargs)
    668 self.tube.scheduler.await_epoch(self._current_epoch)
    669 if self._to_throw:
--> 670     self._throw_error()
    671 self._logger.debug("collecting return")
    672 return self.collect_return(self._current_epoch)

File ~/git/noob/src/noob/runner/zmq.py:728, in ZMQRunner._throw_error(self)
    724 self._logger.debug(
    725     "Deinitializing before throwing error",
    726 )
    727 self.deinit()
--> 728 raise err['err']

E       ValueError: This node just emits errors
E       
E       Error re-raised from node runner process
E       
E       Original traceback:
E       --------------------
E       Traceback (most recent call last):
E         File "noob/runner/zmq.py", line 348, in run
E           value = runner._node.process(*args, **kwargs)
E         File "noob/src/noob/testing/nodes.py", line 86, in error
E           raise ValueError("This node just emits errors")
E       ValueError: This node just emits errors
```

### DIFF
--- a/src/noob/network/message.py
+++ b/src/noob/network/message.py
@@ -4,7 +4,6 @@ import pickle
 import sys
 from datetime import UTC, datetime
 from enum import StrEnum
-from traceback import TracebackException
 from typing import Annotated as A
 from typing import Any, Literal
 
@@ -90,7 +89,7 @@ class AnnounceValue(TypedDict):
 class ErrorValue(TypedDict):
     err_type: type[Exception]
     err_args: tuple
-    traceback: TracebackException
+    traceback: str
 
 
 class AnnounceMsg(Message):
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
reviewed in person

PR #79, opened 2025-12-11
add entrypoint hooks and function to add config sources

### LINKED ISSUE #67: entrypoints and hooks to add config sources
other packages and pretty much every case that someone wants to use noob will need to be able to customize the config source to be able to add more places to find pipeline definitions.

this can be done currently at an end-user level by customizing the user config directory with a `noob_config.yaml` file (or env vars, etc.) but this can only set one directory for a whole environment.

instead we need a way for packages that want to ship pipelines to be able to have those automatically discovered, and that's exactly what [entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html) are for.

we also want a functional way to do this, somethign as simple as an `add_source(path: Path)` function. 

so for an entrypoint we want something like `add_sources() -> list[Path]` functions that can be declared from some other package like

```toml
[project.entry-points."noob.add_sources"]
tubes = "my_package.something:add_sources"
``` 

and then from noob we would do something like

```python
from importlib.metadata import entry_points

class YAMLMixin:
    @classmethod
    def config_sources(cls) -> list[Path]:
        extra_sources = []
        for ext in entry_points(group="noob.add_sources"):
            extra_sources.extend(ext.load()())
        return [Config().config_dir, *extra_sources]
```

and then for the functional hook we would do something like

```python
EXTRA_SOURCES = []

def add_source(path: Path):
    global EXTRA_SOURCES
    EXTRA_SOURCES.append(path)

class YAMLMixin:
    @classmethod
    def config_sources(cls) -> list[Path]:
        global EXTRA_SOURCES

        extra_sources = []
        for ext in entry_points(group="noob.add_sources"):
            extra_sources.extend(ext.load()())
        return [Config().config_dir, *EXTRA_SOURCES, *extra_sources]
```

sneakers-the-rat:
fix: #67 

- [x] function
- [x] entrypoints
- [x] tests
- [x] docs

### DIFF
--- a/src/noob/config.py
+++ b/src/noob/config.py
@@ -1,3 +1,5 @@
+import warnings
+from importlib.metadata import entry_points
 from pathlib import Path
 from typing import Literal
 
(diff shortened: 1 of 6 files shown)


sneakers-the-rat:
oop i pushed this from the wrong branch, accidentally did `gti switch`, but it's fine

raymondwjang (approved):

PR #80, opened 2025-12-11
pylock

### LINKED ISSUE #68: use pylock
there's a standardized lockfile format now, we should use that - https://packaging.python.org/en/latest/specifications/pylock-toml/

https://pdm-project.org/latest/usage/lockfile/#change-lock-file-format

sneakers-the-rat:
fix #68 

swaps out pdm lock with pylock.

not sure yet if the config in the pyproject.toml file actually works or if it's a system config setting, gonna test with CI.

i'm noticing that the pylock doesn't have the optional dependencies in it, so i'm gonna read more tomorrow about the spec or if pdm just hadn't implemented that yet or something. leaving as a draft until then


raymondwjang:
wth...
```sh
[PdmUsageError]: Command 'mypy' is not found in your PATH.
```

raymondwjang (approved):
lgtm

PR #109, opened 2025-12-20
topo sorter takeover

### LINKED ISSUE #78: Take over `graphlib.TopologicalSorter` responsibilities
the builtin graphlib topo sorter has served us well, but we are well and truly beyond what it is supposed to be for, and we now have to do a ton of topo sorter surgery where we access and mutate its private attributes to make it do stuff that it can't do.

topo sort is a super simple algorithm to implement ourselves, and the topo sorter is a super simple class. 

We could either subclass or write our own topo sorter class, but we should do that and resolve the shitty hacks in the scheduler class before we do much else, especially stuff that involves changes to the scheduler.

raymondwjang:
ref: #78 

- [ ] copy/paste graphlib topo sorter tests (that test behaviors that we want to keep) - https://github.com/python/cpython/blob/main/Lib/test/test_graphlib.py

----
📚 Documentation preview 📚: https://noob--109.org.readthedocs.build/en/109/


src/noob/scheduler.py
```diff
@@ -396,13 +490,14 @@ def _init_graph(nodes: dict[str, NodeSpecification], edges: list[Edge]) -> Topol
...
             required_edges = [
-                e.source_node
-                for e in edges
-                if e.target_node == node_id and e.target_node in enabled_nodes
+                e.source_node for e in edges if e.target_node == node_id and e.target_node in nodes
             ]
             sorter.add(node_id, *required_edges)
+        disabled_nodes = [node_id for node_id, node in nodes.items() if not node.enabled]
+        for node_id in disabled_nodes:
+            sorter.remove_node(node_id)
```
sneakers-the-rat: why would we add the nodes and then remove them?

src/noob/scheduler.py
```diff
@@ -18,14 +21,116 @@
...
+
+    def mark_out(self, *nodes: NodeID) -> None:
+        nx.set_node_attributes(self, {node: False for node in nodes}, NodeState.ready)
+        nx.set_node_attributes(self, {node: True for node in nodes}, NodeState.out)
+        nx.set_node_attributes(self, {node: False for node in nodes}, NodeState.done)
+
+    def mark_done(self, *nodes: NodeID) -> None:
+        nx.set_node_attributes(self, {node: False for node in nodes}, NodeState.ready)
+        nx.set_node_attributes(self, {node: False for node in nodes}, NodeState.out)
+        nx.set_node_attributes(self, {node: True for node in nodes}, NodeState.done)
```
sneakers-the-rat: why not a single attribute `state` that can be `"ready"`, `"out"`, or `"done"`

sneakers-the-rat:
got some bad news - for this benchmark: 

```python
from graphlib import TopologicalSorter as GLibTopoSorter

@pytest.mark.parametrize("loaded_tube", ["testing-kitchen-sink"], indirect=True)
@pytest.mark.parametrize("sorter", ["noob", "graphlib"])
def test_topo_sorter(benchmark: BenchmarkFixture, loaded_tube: Tube, sorter: str, monkeypatch: MonkeyPatch) -> None:
    """
    Our TopoSorter should be at least as fast as graphlib's
    """
    if sorter == "graphlib":
        monkeypatch.setattr(scheduler, "TopoSorter", GLibTopoSorter)

    benchmark(lambda: _run_sorter(loaded_tube))

def _run_sorter(tube: Tube) -> None:
    epoch = tube.scheduler.add_epoch()
    sorter = tube.scheduler._epochs[epoch]
    while sorter.is_active():
        ready_nodes = sorter.get_ready()
        for node in ready_nodes:
            sorter.done(node)
```

along with moving the `init_attrs` call into the `prepare` method so that the API matches between this sorter and graphlib

i get this

| Benchmark | Time (best) | Rel. StdDev | Run time | Iters |
| --- | --- | --- | --- | --- |
| test_topo_sorter[noob-testing-kitchen-sink] |   878,975ns |       36.4% |    3.00s |  2,968 |
│ test_topo_sorter[graphlib-testing-kitchen-sink] |     7,918ns |       13.2% |    2.98s | 30,481 |

So it's roughly 110 times slower than graphlib in the best case. it's ok for it to be a little slower, but that's sort of a lot for something that should effectively take no time.

the profiling looks like this 

Screenshot 2026-01-05 at 10 11 33 PM

where the deepcopy (that looks like it can be avoided) is taking up a ton of time, predictably. the other big bar there is finding cycles. 

so 
- total: 3.9s
- deepcopy: .91s
- find cycles: .882s
- set node attributes: .7716s

so those changes account for 2.5s of 3.9s. the other overhead seems to come from seemingly pointless and expensive operations in networkx - like it calls `find_cycles` while setting node attributes? 

idk there has got to be a way to just directly set attributes on a node without going through that proxy method, it's literally flipping a bool, and there has also got to be a way to do cycle detection without it taking forever. if not, i say we go back to just subclassing the graphlib topo sorter and adding methods to it for being able to remove a node and whatever else we need.

src/noob/scheduler.py
```diff
@@ -161,8 +264,7 @@ def sources_finished(self, epoch: int | None = None) -> bool:
 
         graph = self[-1] if epoch is None else self._epochs[epoch]
         return all(
-            graph._node2info[src].npredecessors != _NODE_OUT  # type: ignore[attr-defined]
-            and src not in graph._ready_nodes  # type: ignore[attr-defined]
+            graph.nodes[src][NodeState.out] is not True and src not in graph.ready_nodes
```
sneakers-the-rat: make a `node_ready` method

src/noob/scheduler.py
```diff
@@ -18,14 +21,116 @@
...
+        if cycle:
+            raise nx.NetworkXError(f"Cycle among nodes detected: {cycle}")
+        nx.set_node_attributes(self, {node: True for node in self.source_nodes}, NodeState.ready)
+
+    def is_active(self) -> bool:
+        """
+        Active if there are nodes currently processing or
+        ready to be processed
+        """
+        return any(self.out_nodes or self.ready_nodes)
```
sneakers-the-rat: and also that we have called `prepare()`

src/noob/scheduler.py
```diff
@@ -18,14 +21,116 @@
...
+    def ready_nodes(self) -> set[NodeID]:
+        return {id_ for id_, node in self.nodes.items() if node[NodeState.ready]}
+
+    @property
+    def out_nodes(self) -> set[NodeID]:
+        return {id_ for id_, node in self.nodes.items() if node[NodeState.out]}
+
+    @property
+    def done_nodes(self) -> set[NodeID]:
+        return {id_ for id_, node in self.nodes.items() if node[NodeState.done]}
```
sneakers-the-rat: seems pretty expensive, we aren't really taking advantage of the graph data structure at all here - we should be able to look up new `ready_nodes` when we mark another node as done by looking at its successors and adding them to a set then - when we mark a node as done, we already know the subset of nodes that are made ready by virtue of the edges between them, so we can just directly access them by hashmap lookup and add them to a set, rather than needing to iterate through all nodes, check their attributes, and construct a new object on every access

src/noob/scheduler.py
```diff
@@ -282,10 +377,9 @@ def await_node(self, node_id: NodeID, epoch: int | None = None) -> MetaEvent:
                     "locked between threads."
                 )
 
-            # do a little graphlib surgery to mark just one event as done.
+            # do a little graphlib surgery to mark just one event as out.
             # threadsafe because we are holding the lock that protects graph mutation
-            self._epochs[epoch]._node2info[node_id].npredecessors = _NODE_OUT  # type: ignore[attr-defined]
-            self._epochs[epoch]._ready_nodes.remove(node_id)  # type: ignore[attr-defined]
+            self._epochs[epoch].mark_out(node_id)
```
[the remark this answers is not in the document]
sneakers-the-rat: oops i thought this was still `cancel` from above, this looks correct.

src/noob/scheduler.py
```diff
@@ -396,13 +490,14 @@ def _init_graph(nodes: dict[str, NodeSpecification], edges: list[Edge]) -> Topol
...
-        sorter: TopologicalSorter = TopologicalSorter()
-        enabled_nodes = [node_id for node_id, node in nodes.items() if node.enabled]
-        for node_id in enabled_nodes:
+        sorter: TopoSorter = TopoSorter()
+        for node_id in nodes:
             required_edges = [
-                e.source_node
-                for e in edges
-                if e.target_node == node_id and e.target_node in enabled_nodes
+                e.source_node for e in edges if e.target_node == node_id and e.target_node in nodes
```
sneakers-the-rat:
```suggestion
                e.source_node for e in edges if e.target_node == node_id
```

redundant check, we know `e.target_node` is in `nodes` because `node_id` comes from iterating over `nodes` and we just checked if `e.target_node == node_id`

pyproject.toml
```diff
@@ -13,6 +13,8 @@ dependencies = [
     "platformdirs>=4.3.7",
     'typing-extensions>=4.15.0;python_version<"3.12"',
     "ruamel-yaml>=0.18.16",
+    "networkx>=3.6",
+    "types-networkx>=3.6.1.20251220",
```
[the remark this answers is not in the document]
raymondwjang: oops, right

src/noob/scheduler.py
```diff
@@ -18,14 +21,116 @@
...
+        self.add_edges_from([(pred, node_id) for pred in preds])
+
+    def prepare(self) -> None:
+        """
+        We're allowing cycles when assets are involved,
+        so we make sure there are no cycles among nodes only.
+
+        Source nodes are marked ready.
+        """
+        G_nodes = deepcopy(self)
```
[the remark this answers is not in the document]
raymondwjang: oh, this was here because we allow cycle when `asset` is involved. although in hindsight, it won't actually look like a cycle since its the same node on a _different_ epoch?

src/noob/scheduler.py
raymondwjang: which returns ready nodes?

src/noob/scheduler.py
```diff
@@ -224,12 +326,8 @@ def done(self, epoch: int, node_id: str) -> MetaEvent | None:
                 # in parallel mode, we don't `get_ready` the preceding ready nodes
                 # so we have to manually mark them as "out"
                 # FIXME: so ugly - need to make our own topo sorter
-                if node_id not in self[epoch]._node2info:  # type: ignore[attr-defined]
+                if node_id not in self[epoch].nodes:
                     raise e
-                self[epoch]._node2info[node_id].npredecessors = _NODE_OUT  # type: ignore[attr-defined]
-                self[epoch]._nfinished += 1  # type: ignore[attr-defined]
-                if node_id in self[epoch]._ready_nodes:  # type: ignore[attr-defined]
-                    self[epoch]._ready_nodes.remove(node_id)  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
raymondwjang: i think `self[epoch].done(node_id)` does all the above?

src/noob/scheduler.py
raymondwjang: right, i was a bit wary of doing that because sometimes we were manually setting nodes `ready` but i can just make the `mark_ready` method add the node to the `ready_nodes` set. I also was afraid of the node attributes becoming redundant with the external sets (nodes have `"ready"` attribute but there also exists `_ready_nodes` set)

src/noob/scheduler.py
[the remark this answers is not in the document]
sneakers-the-rat: exactly, ya. that kind of dependency isn't *really* the same as other kinds of dependencies, since we don't take the depended event and then "run" the asset with it. so it is a special case, but i don't think it should be handled by the topo sorter, that's another information leaky case where we are making classes with specific roles "know too much" about the rest of the system. that sounds like something to have as a contextmanager/wrapper thing around the process method - when an event is yielded that the asset store is depending on, handle that in the runner, which is the thing that orchestrates between the parts of the system.

src/noob/scheduler.py
sneakers-the-rat: we don't need to enforce it, there isn't really a language mechanism for enforcing anything in python :). within the code, we're the only ones touching the object, and we are using an enum - if someone downstream modifies the value to something that isnt a known value and the program has undefined behavior, that's a them problem

src/noob/scheduler.py
sneakers-the-rat: like this one - given some node id, check if the node is ready: https://github.com/miniscope/noob/blob/2af440f1af9a4030e8a6658896fea2d0052dfa72/src/noob/scheduler.py#L126

i say this because there's the possibility for the check for readiness to be done differently in multiple places - so making it a method is a way of making sure there is only one definition of "readiness." here's it's defined as `out` == `False` and not being in `ready_nodes` - we might be tempted to just check if something is in `ready_nodes` elsewhere, etc.

raymondwjang:
the benchmarking results... are disappointing. altho i don't see anywhere in the `nx.set_node_attributes` that calls `find_cycles`? i'll look more into it but that's crazy how slow that is

src/noob/scheduler.py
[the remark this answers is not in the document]
sneakers-the-rat: ah right, my bad - that would work for a node that wasn't previously marked as out.

src/noob/scheduler.py
sneakers-the-rat: also looking now i see this same check is in the current source. it's probably not that expensive since it's just testing list membership, but also no point in having it anyway.

sneakers-the-rat:
i could go either way on using nx - 

on the one hand, i would really like to keep deps as low as we possibly can so this package is easy to include in other packages (e.g. notice how every time we are adding a dependency, i am going to check its dep tree, and being very particular about not adding deps that have heavy trees - so i assume most other package maintainers do the same thing). What we need as far as graph analysis is really minimal as well - we pretty much just need a topo sorter that we add a few additional manipulations to, we don't really need any of the fancier stuff that having a proper graph data model would give us. So it's true that nx probably brings a certain amount of unavoidable overhead just because its models are more complete and capable than a purpose-built class for topo sorting.

on the other hand, maybe we do eventually want to have more proper graph-like things: e.g. having labels on edges would allow us to do stuff like declare dependencies between things that don't result in events being passed between them (e.g. if we wanted some other way of declaring execution order, or if we wanted to have "one big graph" where nodes are labeled by epoch rather than separate graphs for separate epochs). We might be able to use some of the graph algos to be smart about scheduling, like detecting communities or disconnected components in graphs and using that to split runner processes, etc.

I don't have a hard threshold for perf here, and we haven't done a ton of optimization throughout the package, but since this is a core routine that's likely to have some of the highest call counts in the whole package, i think a 100x perf decline is probably not a great precedent to set, and it's worth avoiding it now rather than trying to optimize it out later. i'll leave it up to you

sneakers-the-rat:
in any case the changes here are making the benchmarking tests take so long that they error out from pytest-timeout, so we should raise the timeout in the bench module so we can actually see how big the impact is lol

raymondwjang:
i'm thinking the possible features that comes from `networkx`'s high-level graph algos don't exactly appear to be a part of near dev roadmap. 

the main motivations currently are: 
1. we want to stop having surgeries on toposort
2. support adding nodes / edges _while_ the graph is running - because when assets enter dependency, nodes that use asset in subsequent epoch should not execute before the last node that uses the asset in the previous one is complete.

1 seems easier to address by subclassing toposort, but 2 seems more ugly. toposort does not know how to add nodes / edges after `.prepare`, so `scheduler` would need to retain its `epoch_idx: graph` format and make a whole epoch graph, and then for every process call check if a ready node depends on `asset`, and compare it against the completion status of the previous epoch's last asset-using node.

sneakers-the-rat:
another thing - we might get checking for cycles for free by doing what networkx does: 
- its `is_directed_acyclic_graph` is just `is_directed(g) and not has_cycle(g)`
- `has_cycle` is literally just topo sorting and returning true if `topological_sort` raises an error
- `topological_sort` just yields from `topological_generations`
- `topological_generations` is just Kahn's algorithm, which yields... the topo sort generations, and raises the error if there are any nodes with >1 in degree after the nodes with 0 indegree run out.

so... if instead we just put cycle checking as a pydantic validator on `Tube`, and we just lazily raise errors on encountering cycles then it's win win win: 
- don't have to do cycle checking while running explicitly ever, and that's more than half of the time taken in the profiling gone for free
- we can mutate the graph while we're running without needing to check for cycles every time
- user is alerted before doing any work initializing nodes that the tube is invalid

src/noob/toposort.py
```diff
@@ -0,0 +1,207 @@
...
+        If called multiple times with the same node argument, the set of dependencies
+        will be the union of all dependencies passed in.
+
+        It is possible to add a node with no dependencies (*predecessors* is not provided)
+        as well as provide a dependency twice. If a node that has not been provided before
+        is included among *predecessors* it will be automatically added to the graph with
+        no predecessors of its own.
+        """
+        # Create the node -> predecessor edges
+        nodeinfo = self._get_nodeinfo(node)
```
sneakers-the-rat: refuse to add to a node that is out or completed

src/noob/toposort.py
```diff
@@ -0,0 +1,207 @@
...
+        nodeinfo.npredecessors += len(predecessors)
+        if nodeinfo.npredecessors == 0:
+            self.mark_ready(node)
+        else:
+            # in case node is called multiple times
+            self._ready_nodes.discard(node)
```
sneakers-the-rat: not just adding the length of the precessors, we need to check if the predecessors have already been completed - if they are already done, we don't increment the npredecessors

src/noob/toposort.py
```diff
@@ -0,0 +1,207 @@
...
+        """Return ``True`` if more progress can be made and ``False`` otherwise.
+
+        Progress can be made if cycles do not block the resolution and either there
+        are still nodes ready that haven't yet been returned by "get_ready" or the
+        number of nodes marked "done" is less than the number that have been returned
+        by "get_ready".
+
+        Raises ValueError if called without calling "prepare" previously.
+        """
+        return self._nfinished < self._npassedout or bool(self.ready_nodes)
```
sneakers-the-rat: do the passive check for cycles here - if `ready_nodes == {} and (nfinished + npassedout < len(self._node2info)` then we have cycles

sneakers-the-rat (approved):
good with this except changes as discussed and tests that need to be written. will push a set of stub tests shortly

src/noob/scheduler.py
```diff
@@ -396,13 +377,14 @@ def _init_graph(nodes: dict[str, NodeSpecification], edges: list[Edge]) -> Topol
...
+
+    def assert_acyclic(self) -> None:
+        """
+        Checks that the graph is acyclic.
+        """
+        graph = self._init_graph(nodes=self.nodes, edges=self.edges)
+        cycle = graph.find_cycle()
+        if cycle:
+            raise AssertionError("nodes are in a cycle", cycle)
+        return None
```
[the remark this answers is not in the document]
raymondwjang: done

src/noob/toposort.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # List of successor nodes. The list can contain duplicated elements as
+        # long as they're all reflected in the successor's npredecessors attribute.
+        self.successors: list[NodeID] = []
+
+
+class TopoSorter:
+    """
+    Provides functionality to topologically sort a graph of `class: .node.base.Node`.
+    """
```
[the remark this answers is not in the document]
raymondwjang: done

src/noob/toposort.py
raymondwjang:
```python
        # Refuse to add nodes that are out / done
        reject = {self.out_nodes: "already out", self.done_nodes: "already done"}
        reasons = [reason for group, reason in reject.items() if node in group]
        if reasons:
            raise ValueError(f"{node} cannot be added: {", ".join(reasons)}.")
```

src/noob/scheduler.py
```diff
@@ -102,6 +104,18 @@ def get_ready(self, epoch: int | None = None) -> list[MetaEvent]:
 
         graphs = self._epochs.items() if epoch is None else [(epoch, self._epochs[epoch])]
 
+        maybe_ready = set()
```
raymondwjang: @sneakers-the-rat checking your feelings about this part (handling `input` and `assets` nodes in scheduler instead of runner)

tests/data/pipelines/cycle.yaml
```diff
@@ -0,0 +1,14 @@
...
+  a:
+    type: noob.testing.count_source
+    depends: c.value
+  b:
+    type: noob.testing.multiply
+    depends:
+      - left: a.index
+  c:
+    type: return
+    depends: b.value
```
sneakers-the-rat: return is a special case - let's not use special cases when testing features that don't use the special case

src/noob/toposort.py
```diff
@@ -0,0 +1,242 @@
...
+        self._ready_nodes.update(nodes)
+
+    def mark_out(self, *nodes: NodeID) -> None:
+        for node in nodes:
+            # there isn't a good way to discard multiple elements in bulk
+            self._ready_nodes.discard(node)  # missing ok
+        self._out_nodes.update(nodes)
+        self._npassedout += len(nodes)
+
+    def mark_expired(self, *nodes: NodeID) -> None:
```
sneakers-the-rat: if we call it expired here, then we should also call it `Scheduler.expire` rather than `Scheduler.cancel`

src/noob/toposort.py
```diff
@@ -0,0 +1,242 @@
...
+        return self._out_nodes
+
+    @property
+    def done_nodes(self) -> set[NodeID]:
+        return self._done_nodes
+
+    def mark_ready(self, *nodes: NodeID) -> None:
+        self._ready_nodes.update(nodes)
+
+    def mark_out(self, *nodes: NodeID) -> None:
```
sneakers-the-rat: docstring

src/noob/toposort.py
```diff
@@ -0,0 +1,242 @@
...
+        for node in nodes:
+            # there isn't a good way to discard multiple elements in bulk
```
sneakers-the-rat: set difference
```
self._ready_nodes = self._ready_nodes - set(nodes)
```

src/noob/toposort.py
```diff
@@ -0,0 +1,242 @@
...
+    __slots__ = "node", "nqueue", "successors"
+
+    def __init__(self, node: str) -> None:
+        # The node this class is augmenting.
+        self.node = node
+
+        # Number of predecessors, generally >= 0. When this value falls to 0,
+        # and is returned by get_ready(), this is set to _NODE_OUT and when the
+        # node is marked done by a call to done(), set to _NODE_DONE.
+        self.nqueue = 0
```
sneakers-the-rat: name is a bit less clear to me than `n_predecessors` but we've gone back and forth enough already on it so fine

src/noob/toposort.py
```diff
@@ -0,0 +1,258 @@
...
+        # i.e., we filter edges to any node that are explicitly disabled, but pass others.
+        disabled_nodes = [node_id for node_id, node in nodes.items() if not node.enabled]
+        for e in edges:
+            if e.target_node in disabled_nodes:
+                continue
+            self.add(e.target_node, e.source_node)
+        # add enabled nodes that have no edges
+        for node_id, node in nodes.items():
+            if node.enabled and node_id not in self._node2info:
+                self.add(node_id)
```
sneakers-the-rat: made this change to allow empty initialization, as well as edge-only initialization - mostly for the graphlib tests, but also makes sense to be able to do this.

sneakers-the-rat (commented):
alright cool, final changes.

src/noob/toposort.py
sneakers-the-rat: https://github.com/miniscope/noob/pull/109/commits/f56103388662cf1c5b7c0225bcddef47d8d9a02d

src/noob/toposort.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def out_nodes(self) -> set[NodeID]:
+        return self._out_nodes
+
+    @property
+    def done_nodes(self) -> set[NodeID]:
+        return self._done_nodes
+
+    def mark_ready(self, *nodes: NodeID) -> None:
```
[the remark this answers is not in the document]
sneakers-the-rat: https://github.com/miniscope/noob/pull/109/commits/f56103388662cf1c5b7c0225bcddef47d8d9a02d

src/noob/toposort.py
sneakers-the-rat: https://github.com/miniscope/noob/pull/109/commits/f56103388662cf1c5b7c0225bcddef47d8d9a02d

src/noob/scheduler.py
```diff
@@ -107,6 +104,18 @@ def get_ready(self, epoch: int | None = None) -> list[MetaEvent]:
...
+        special_nodes = True
+        while special_nodes:
+            special_nodes = False
+            for epoch, graph in graphs:
+                for node_id in graph.get_ready():
+                    if node_id in SPECIAL_NODES:
+                        self.done(epoch, node_id)
+                        special_nodes = True
+                    else:
+                        maybe_ready.add((epoch, node_id))
```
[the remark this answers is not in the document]
raymondwjang: i keep limiting my thoughts to sync runner. good thing we discussed this

src/noob/scheduler.py
```diff
@@ -15,8 +14,11 @@
 from noob.exceptions import EpochCompletedError, EpochExistsError
 from noob.logging import init_logger
 from noob.node import Edge, NodeSpecification
+from noob.toposort import TopoSorter
 from noob.types import NodeID
 
+SPECIAL_NODES = ("input", "assets")
```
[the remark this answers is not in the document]
raymondwjang: in which case i think we can remove this from scheduler

src/noob/toposort.py
```diff
@@ -0,0 +1,258 @@
...
+        if reasons:
+            raise ValueError(f"{node} cannot be added: {', '.join(reasons)}")
+
+        # Create the predecessor -> node edges
+        # filter predecessors to only those that are newly being created
+        new_predecessors = []
+        for pred in predecessors:
+            pred_info = self._get_nodeinfo(pred)
+            if node in pred_info.successors:
+                continue
```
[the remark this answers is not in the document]
raymondwjang: sounds good

sneakers-the-rat (approved):

Issue #23, opened 2025-06-30
Convert `SynchronousRunner.process` spaghetti into lifecycle events


sneakers-the-rat:
The logic of the `runner.process` method is currently spaghetti that looks like this:

```python
        graph = self.tube.graph()
        graph.prepare()

        while graph.is_active():
            for node_id in graph.get_ready():
                node = self.tube.nodes[node_id]
                node_input = self.gather_input(node)
                if node_input is None:
                    graph.done(node_id)
                    self._logger.debug(f"Node {node_id} received no input, skipping")
                    continue
                value = node.process(**node_input)
                self.store.add(value, node_id)
                graph.done(node_id)
                self._logger.debug(f"Node {node_id} emitted %s", value)
```

we should instead split that up into different phases of processing so that different runners can implement them differently. this should simplify implementation as well as keep ourselves sane.

something like this ( this is just pseudocode example illustrating the point that these should all be methods on the runner class)

```python
graph = self.prepare_graph()
while graph.is_active():
    node = self.get_node(node_id)
    args, kwargs = self.gather_input(node)
    events = self.call_node(node, args, kwargs)
    self.store_events(events)
    self.update_graph(node, events)
```

so that then we have a specific lifecycle of operations that happen within `process` and are not tempted to keep making a bunch of special cases and stuff for the sync runner and make it impossible to duplicate in other runners when it's time to do them.

PR #119, opened 2026-01-07
calling async from sync in the most cursed way

### LINKED ISSUE #42: Handle `async` process methods in `sync` runners
Async running functionality needed for operations such as video encoding.

```python
    @process_method
    def save(self, frame: Frame) -> None:
        frame = frame.array.astype(np.uint8)
        self._stream.width = frame.sizes[AXIS.width_dim]
        self._stream.height = frame.sizes[AXIS.height_dim]

        try:
            vid_frame = av.VideoFrame.from_ndarray(frame.to_numpy(), format="gray")
            packets = self._stream.encode(vid_frame.reformat(format=self._stream.pix_fmt))  # to a separate thread
            for packet in packets:
                self._container.mux(packet)

        except EncodingWarning as e:
            self._container.close()
            raise EncodingError() from e
```

sneakers-the-rat:
Fix: #42 

Punting docs to #120 

OK so this was was harder than i thought. The strategy for calling async from sync is mostly described in the docstring, but i'll reproduce that here:

> Call an async function synchronously, either in this thread or a subthread.
> 
> So here's the deal with this nonsense:
> 
> * Calling async from sync is easy when there is no running eventloop in the thread
> * Calling async from sync is **almost comically hard** when there is a running eventloop.
> 
> We are likely to encounter the second case where, e.g.,
> some async application calls some other code that uses a :class:`.SyncRunner`
> to run a tube that has an async node.
> 
> :func:`asyncio.run` and :class:`asyncio.Runner` refuse to run when there is live eventloop,
> and attempting to use any of the :class:`~asyncio.Task` or :class:`~asyncio.Future`
> spawning methods from the running eventloop like :meth:`~asyncio.AbstractEventLoop.call_soon`
> or :func:`asyncio.run_coroutine_threadsafe`
> and then polling for the result with :meth:`asyncio.Future.result`
> causes a deadlock between the outer sync thread and the eventloop.
> The basic problem is that there is no way to wait in the thread (synchronously)
> that yields the thread to the eventloop (which is what async functions are for).
> 
> We need to make a new thread in **some** way,
> Django's ``asgiref`` has a mindblowingly complicated
> `async_to_sync <https://github.com/django/asgiref/blob/2b28409ab83b3e4cf6fed9019403b71f8d7d1c51/asgiref/sync.py#L585>`
> function that works **roughly** by creating a new thread
> and then calling :func:`asyncio.run` from *within that*
> (plus about a thousand other things to manage all the edge cases).
> That's more than a little bit cursed, because ideally,
> since the hard case here is where there is already an eventloop in the outer thread,
> we would be able to just *use that eventloop*.
> Normally one would just ``await`` the coro directly,
> which is what :class:`.AsyncRunner` does,
> but the :class:`.SyncRunner` can't do that because :meth:`.SyncRunner.process` is sync.
> 
> However if one creates a new thread with a new eventloop,
> that will break any stateful nodes that e.g. have objects like :class:`asyncio.Event`
> that are bound to the first eventloop.
> 
> Until we can figure out how to reuse the outer eventloop,
> we do the best we can with a modified version of ``asgiref`` 's approach.
> 
> * Create a :class:`asyncio.Future` to store the eventual result we will return
>   (the "result future")
> * Wrap the coroutine to call in another coroutine that calls :meth:`~asyncio.Future.set_result`
>   or :meth:`~asyncio.Future.set_exception` on some passed future
>   rather than returning the result directly
> * Use a :class:`~concurrent.futures.ThreadPoolExecutor` to run the wrapped coroutine
>   in a **new** :class:`asyncio.AbstractEventLoop` in a separate thread,
>   returning a second :class:`~concurrent.futures.Future` (the "completion future")
> * Add a callback to the completion future that notifies a :class:`~threading.Condition`
> * Wait in the main thread for the :class:`~threading.Condition` to be notified
> * Return the result from the result future.
> 
> The reason we don't just directly return the value of the process coroutine
> in the inner wrapper coroutine
> and then return the result of the completion future is error handling -
> Errors raised in the wrapping coroutine have a large amount of noise in the traceback,
> so instead we use :meth:`~asyncio.Future.set_exception` to propagate the raised error
> up to the main thread and raise it there.

for reference: https://github.com/django/asgiref/blob/2b28409ab83b3e4cf6fed9019403b71f8d7d1c51/asgiref/sync.py#L585

In testing, discovered some other problems:
- Async runner was not handling exceptions raised within nodes, as they were raised in the `node_complete` callback, which didn't kill the `process` method like i thought it would (idk why i thought it would). so as a result the asyncrunner would just wait indefinitely for the node ready event to be set. Added explicit error handling
- ZMQRunner (specifically the EventLoopMixin part) could not run when there was already an eventloop running outside, as it would try and use that loop instead of creating one within the thread that it spawns. so forced it to use the loop from within the thread, start that before creating the sockets, and all is well.

But now! 
- SyncRunner can run async nodes
- ZMQRunner can run async nodes

Benchmarking the difference between spawning a thread on demand and using a premade ThreadPoolExecutor:
- on demand: 413us (best), 3212 iters
- premade: 322us (best), 4088 iters

so 1.3 times as long at worst and on average. not *too* terrible, but i'm not going to handle setting up and tearing down the thread pool executor until i do #23 so i can do it cleanly.

----
📚 Documentation preview 📚: https://noob--119.org.readthedocs.build/en/119/

### DIFF
--- a/src/noob/runner/asyncio.py
+++ b/src/noob/runner/asyncio.py
@@ -20,6 +20,11 @@ else:
 class AsyncRunner(TubeRunner):
 
     eventloop: asyncio.AbstractEventLoop | None = None
+    exception_timeout: float = 10
+    """
+    When a node raises an error, wait this long (in seconds)
+    before cancelling the other currently running nodes. 
+    """
 
     def __post_init__(self):
         super().__post_init__()
@@ -27,6 +32,7 @@ class AsyncRunner(TubeRunner):
         self._node_ready = asyncio.Event()
         self._init_lock = asyncio.Lock()
         self._pending_futures = set()
+        self._exception: BaseException | None = None
         if not self.eventloop or self.eventloop.is_closed():
             self.eventloop = asyncio.get_running_loop()
 
@@ -68,6 +74,17 @@ class AsyncRunner(TubeRunner):
         calling their process method and passing events as they are emitted.
 
         Process-scoped ``input`` s can be passed as kwargs.
+
+        We don't want to cancel nodes on the first error,
+        as would happen with a task group.
+        e.g. some nodes might be writing data, and we want that to succeed
+        even if some other node in this generation fails.
+        We also want to schedule nodes opportunistically,
+        as soon as their dependencies are satisfied,
+        without needing to wait for the entire graph generation to finish
+        so we queue as many as we can and use futures/callbacks to signal completion.
+        If a node raises an error, we allow the other running nodes to complete
+        (with some timeout
         """
         self.eventloop = cast(asyncio.AbstractEventLoop, self.eventloop)
         if not self._running.is_set():
@@ -80,21 +97,14 @@ class AsyncRunner(TubeRunner):
         scheduler.add_epoch()
 
         while scheduler.is_active():
+            if self._exception:
+                await self._raise_exception()
             ready = scheduler.get_ready()
             if not ready:
                 # if none are ready, wait until another node is complete and check again
                 self._node_ready.clear()
                 await self._node_ready.wait()
                 continue
-
-            # we don't want to cancel nodes on the first error,
-            # as would happen with a task group.
-            # e.g. some nodes might be writing data, and we want that to succeed
-            # even if some other node in this generation fails.
-            # We also want to schedule nodes opportunistically,
-            # as soon as their dependencies are satisfied,
-            # without needing to wait for the entire graph generation to finish
-            # so we queue as many as we can and use futures/callbacks to signal completion
 
             for node_info in ready:
                 node_id, epoch = node_info["value"], node_info["epoch"]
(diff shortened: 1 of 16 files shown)


raymondwjang:
> The reason we don't just directly return the value of the process coroutine
in the inner wrapper coroutine
and then return the result of the completion future is error handling -
Errors raised in the wrapping coroutine have a large amount of noise in the traceback,
so instead we use :meth:~asyncio.Future.set_exception to propagate the raised error
up to the main thread and raise it there.

need to take time to digest these better lol

src/noob/runner/asyncio.py
```diff
@@ -20,13 +20,19 @@
 class AsyncRunner(TubeRunner):
 
     eventloop: asyncio.AbstractEventLoop | None = None
+    exception_timeout: float = 10
```
raymondwjang: how come we're waiting to cancel?

sneakers-the-rat: Because other nodes are running, and we dont want to just kill them if we can just wait and theyll complete naturally - e.g. if something is doing file i/o we dont want to corrupt the file just because some other node fails.

raymondwjang (approved):

PR #122, opened 2026-01-08
inputs in zmq runner

### LINKED ISSUE #70: Handle inputs for zmq runner
#54 handles the most basic version of zmq runner, but we need to implement `input`s for it! 

just add the inputs to the `Process` message!

sneakers-the-rat:
Fix: #70 

i think pretty self explanatory?

We give tube-scoped inputs to the node classes when creating them, then we give process-scoped inputs in the `ProcessMessage`. Since we can have potentially multiple epochs in queue and also the process calls can come out of order, we don't use the input collection like we do in the sync runner, but instead use the event store. this all ends up working pretty much perfectly between the eventstore and scheduler - `inputs` are dependencies in the scheduler that we mark done when we get them, then we get their values from the eventstore like we would any other events...

then i just made the tests parameterized over both runners, and since those are the same tests we have for input generally, zmq is at feature parity for inputs.

----
📚 Documentation preview 📚: https://noob--122.org.readthedocs.build/en/122/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,6 +138,7 @@ markers = [
     "async_runner: uses the async runner",
     "sync_runner: uses the sync runner",
     "zmq_runner: uses the zmq runner",
+    "input: tube and process inputs",
 ]
 
 [tool.black]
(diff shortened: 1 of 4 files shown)


sneakers-the-rat:
@raymondwjang gave approval verbally like some kind of high power executive trying to make it to their next meeting

PR #125, opened 2026-01-13
Raise when trying to use `run` or `iter` on tubes with process-scoped inputs

### LINKED ISSUE #124: Raise error when attempting to call `run` when there are `process` scoped inputs
It's invalid to call `run` when there are `process` scoped inputs, because you can't just keep running epochs without having the inputs! 

The only exception to this would be maybe for `iter` where you want to run some fixed number of iterations and provide the inputs for all of those

This should raise an error basically - if you try and call run, check if inputs with a process scope, and yell about it.

sneakers-the-rat:
Fix: #124

----
📚 Documentation preview 📚: https://noob--125.org.readthedocs.build/en/125/

### DIFF
--- a/src/noob/runner/base.py
+++ b/src/noob/runner/base.py
@@ -16,6 +16,8 @@ from typing import Any, ParamSpec, Self, TypeVar
 
 from noob import Tube, init_logger
 from noob.event import Event, MetaEvent
+from noob.exceptions import InputMissingError
+from noob.input import InputScope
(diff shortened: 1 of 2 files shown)


sneakers-the-rat:
review after #127 , i think the status announce fix there rescues some of the async problems

raymondwjang (approved):
short and sweet, lgtm

PR #127, opened 2026-01-13
Node statefulness

### LINKED ISSUE #123: Statefulness for nodes
We have a general issue in concurrent runners - only the ZMQ runner for now - where once we have a free running mode where each run is not bound to a particular process call then it's possible (and likely) that events will occur out of order, or a given nodes dependencies will not necessarily be satisfied in epoch order. 

That is a problem if a node is stateful: e.g. A node that is collecting parts of an image frame needs to get those parts in order or the image will be all messed up. 

However for nodes that are *not* stateful, we want to be able to take advantage of that and run epochs whenever our dependencies are fulfilled, as soon as we can run them.

So we need some property `stateful`, but how and where it should be applied is 5% tricky given that we want to be wrapping arbitrary classes and functions. In general, statefulness is inherent to the actual node class itself (rather than its application in a given tube), but we may not be able to declare that on the node class itself (since we might not own it/it doesn't know about noob at all/etc.)

So i think we need to be able to specify it in two places:
- On the node class itself (and function nodes will probably need to gain some decorator as well, more on that blow)
- In the tube specification

Where the tube specification overrides the node class declaration, if set.

To many this not extremely annoying, we make the conservative assumption on defaults:
- Class nodes default to stateful
- Generators default to stateful (though currently they can only be used as sources)
- Functions default to stateless.

There are some functions, however, that might want to declare themselves as stateful if, e.g. they refer to some module global or something like that, so we may want to have a decorator that lets people declare a function node as stateful.

As far as implementation goes, this *should be* a class attribute, since it applies to the class not the instance, but since it's possible to vary it also needs some instance counterpart. Since the syntax is the same in this case (subclasses would just set `stateful = False` in their class body as a default)  I think we can just keep it as an instance attr

sneakers-the-rat:
Fix: #123 

disturbingly simple. 

most of this was already in place:
- inputs are stored along with their epoch
- scheduler await_node can wait for a node being ready in a particular epoch
- we track how many times `process()` has been called so we can catch up

so the actual functional part of the feature is literally two lines.

the rest of it is adding the `stateful` flag to nodes and node specs - i made it an instance attribute rather than a class var so it would be straightforward to override from the spec. 

----
📚 Documentation preview 📚: https://noob--127.org.readthedocs.build/en/127/


sneakers-the-rat:
i think the debug logs being so goddamn expensive might be causing a lot of problem here, so i made them a flag

src/noob/runner/zmq.py
```diff
@@ -408,13 +408,17 @@ def await_inputs(self) -> Generator[tuple[list[Any], dict[str, Any], int]]:
...
             if not self._freerun.is_set():
-                if self._to_process == 0:
+                if self._to_process <= 0:
+                    self._to_process = 0
                     self._process_one.wait()
                 self._to_process -= 1
-                if self._to_process == 0:
+                if self._to_process <= 0:
+                    self._to_process = 0
                     self._process_one.clear()
```
sneakers-the-rat: was possible to go below zero and run infinitely

src/noob/node/base.py
```diff
@@ -238,17 +269,24 @@ def from_specification(
             ):
                 raise ValueError("No input collection supplied, but inputs specified in params")
 
+        # additional kwargs that can be present or absent without default
+        kwargs = {}
```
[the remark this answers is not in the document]
sneakers-the-rat: yeah - there are variable defaults, so we can't do a shared default like spec.enabled. we need to know not just true or false, but whether the parameter was explicitly provided or not. the alternative to this is to add a validator to the ClassWrap node that mimics the one on the function wrap node or have switching logic outside the node classes (e.g. below when detecting what wrap class to use).

this seemed simpler, since this is the one interface between nodes and node specs, and the one place it would be relevant.

tests/test_runners/test_zmq.py
```diff
@@ -59,3 +64,130 @@ def test_error_reporting():
...
+    runner = ZMQRunner(tube=tube)
+    with runner:
+        runner.command.add_callback("inbox", _event_cb)
+        # skip first epoch
+        runner.command.process(1, input={"multiply": 3})
+
+        start = time()
+        while len(events) < 1 and time() - start < 1:
+            await sleep(0.1)
+        # # we should have only received an event from the stateless count source
```
raymondwjang: why would a and b not generate events?

sneakers-the-rat: they are stateful. they depend on input from epoch 0, and will not run if they receive inputs from later epochs until they get zero. that's sorta the whole PR. if they emitted events, meaning they called `process()` with input from epoch 1, then that means they are not being treated as stateful.

src/noob/runner/zmq.py
```diff
@@ -577,11 +585,11 @@ def on_announce(self, msg: AnnounceMsg) -> None:
...
             # status and announce messages can be received out of order,
             # so if we observe the command node being out of sync, we update it.
-            if (
+            elif (
                 self._node.id in msg.value["nodes"]
-                and msg.value["nodes"][self._node.id]["status"] != self.status
+                and msg.value["nodes"][self._node.id]["status"] != self.status.value
             ):
-                self.identify()
+                self.update_status(self.status)
```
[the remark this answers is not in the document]
sneakers-the-rat: infinite annouce loop

tests/test_runners/test_zmq.py
raymondwjang: ohh the epoch is 1

raymondwjang: got it makes sense

raymondwjang (approved):

PR #128, opened 2026-01-13
Declutter process, convert to hook methods

sneakers-the-rat:
Fix: #23 

This is sort of half tidiness and half extensibility - we end up with a ton of repeated code in places that only require minor variation becasue process is such a huge monolith. this also splits it up into logical sections so that we can keep track of all the steps that need to be done, so we don't forget them or have divergent behavior.

so this is a draft of cleaning it up by splitting it up into hook methods. Ideally we would be able to make the top level `TubeRunner.process` just be the body of what's in the sync runner now if we can make it general enough. then subclasses just override the different parts of the process that they want to change.

opening this as a draft to check what you think before refactoring the asyncio runner. not sure how much will be applicable to the zmqrunner because it's so different.

most of this is just reorganization, but i did fix one problem with the scheduler - it couldn't *quite* handle disabled nodes, so the place where the syncrunner went like `if not ready: scheduler.end_epoch()` was load bearing and defies intuition about how enabled/disabled nodes work. the changes make it so if we encounter a disabled node either during creation or during running we don't end up passing it out or waiting on it, but we don't just remove it from the graph so that nodes that depend on it run when they need that node.

---

After #130 made a few updates too:
- Add contextmanager for asset state
- Selectively initialize node scoped assets so they only initialize when they should

----
📚 Documentation preview 📚: https://noob--128.org.readthedocs.build/en/128/


sneakers-the-rat:
i think this codspeed benchmarking is probably too variable to be informative.. like this PR just reorganizes the way that the process method is called but it's otherwise pretty much exactly the same. the only thing that could be degrading perf are the changes to the topo sorter to handle disabling nodes, but that isn't even tested in the benchmark that's failing because it's too slow, and we should see it in the benchmark that specifically tests topo sorter

sneakers-the-rat:
ok since we're removing xarray in #130 already, i did the same here because importing it was making the `simulation` instrumentation of codspeed fail. theoretically the cpu simulation instrument should be less variable than the walltime instrument? anyway this PR should probably happen after #130 so that I can incorporate the changes w.r.t. the asset initialization there.

src/noob/runner/base.py
```diff
@@ -199,6 +215,24 @@ def running(self) -> bool:
         """
         pass
 
+    def _process_node(self, node_info: MetaEvent, input: dict) -> None:
```
raymondwjang: wow this is beautiful

raymondwjang (approved):
lgtm more codes but becoming cleaner and easier to understand i'm in awe

Issue #77, opened 2025-12-11
`runner` scoped Assets can have `depends` to be able to structure mutations - "virtual cycles"


sneakers-the-rat:
Problem: we have assets, and in cala they are used to avoid copying lots of big memory things. the problem is that the graph structure of mutations of those assets is implicit: it's only through other dependencies between nodes that we know that they shouldn't be run at the same time (and potentially mutate the asset at the same time). This is also a problem for multiprocessing runners because we don't have a good way of handling data locality or knowing what needs to be sent between each node. 

so we need a way to structure the graph of operations around an asset, and the natural way to do that would be to just use dependencies: a 'first node' depends on the asset directly, and then it emits it as an event. successor nodes that want the asset depend on that prior node's output, and so on.

the problem then is that we can't 'reuse' the asset between `process` calls because it's not clear where it ends up: as discussed in #60 , we can only reliably tell that a returned value from a node is a mutated asset *if the `id()` remains unchanged* - i.e. that the object was changed in place rather than consumed and re-created with something that should take its place. We could allow that to be annotated in node signatures, but that's another issue. If we could reliably detect the last place an asset comes out, we could just 'automatically' use that as the state of the asset to be given to nodes that need it in the next round of `process`,  but that seems error prone and somewhat implicit.

instead, we could allow `runner` scoped assets to `depend` on one output of a node in a pipeline, so then
- on `init` the asset is created, and on the first run the first node to depend on it gets the fresh version
- when the depended on node emits the relevant signal, that gets added back into the `Store`
- on the next epoch, that version from the store is given to the node.

To do this, we also need to write some way of having inter-epoch dependencies, which also seems like a separate issue. The "first node" that receives the asset needs to know that it can't run until the last node that the asset depends on runs *in the previous epoch*.

the caveat with this is that it's on the user to ensure that they don't structure their graph such that the asset would be received and mutated by two nodes in the same 'layer' of the topo sort graph, which feels fine to me - e.g. someone could have a chain of nodes 1-2-3 that each receive and pass the asset in series, but the asset depends on the output of 2 and not 3. this would be something we would *want* to allow if 3 were just reading the asset and not mutating it, because then the following process cycle could start as soon as `2` was done.

raymondwjang:
so, a few things to consider:
1. i cannot necessarily stop the user from mutating an asset in-place. if they do that and create some race conditions, that's on them
[...]

PR #130, opened 2026-01-13
Runner scoped assets can have depends to be able to structure mutations   virtual cycles

raymondwjang:
ref: #77 

----
📚 Documentation preview 📚: https://noob--130.org.readthedocs.build/en/130/

### DIFF
--- a/src/noob/testing/nodes.py
+++ b/src/noob/testing/nodes.py
@@ -207,7 +207,7 @@ def jump(generator: count, n: int = 1) -> tuple[A[count, Name("skirttt")], A[int
 
 
 def rewind(generator: count, n: int = 1) -> A[count, Name("skrittt")]:
-    """Purposely designed to diverge inplace change and output"""
+    """Purposely designed to mutate the input and return a new object"""
     return count(next(generator) - n)
 
 


src/noob/asset.py
```diff
@@ -48,7 +48,7 @@ class AssetSpecification(BaseModel):
     """The scope of the asset. See :class:`.AssetScope`"""
     params: dict | None = None
     """Initialization parameters"""
-    depends: list[AbsoluteIdentifier] | None = None
+    depends: AbsoluteIdentifier | None = None
```
sneakers-the-rat:
```suggestion
    depends: DependencyIdentifier | None = None
```

src/noob/tube.py
```diff
@@ -216,9 +216,11 @@ def from_specification(
         # adding the input validates presence of required inputs
         input_collection.add_input(InputScope.tube, input)
 
+        state = cls._init_state(spec)
+        state.init_assets(AssetScope.tube)
```
sneakers-the-rat: i think this should happen in the runner init methods: maybe our naming is a little confusing here, but the the `_init_nodes` and etc. methods just create the objects, but defer calling their `init()` methods until it's time to actually run the thing. Let's keep that pattern consistent: everything has the same set of states:

- Instantiated: object exists, but not prepared
- Initialized: Ready to run
- Running: currently in the middle of a `process()` call, or otherwise actively being used


where we have 
- instantiated -> initialized: `init()`
- initialized -> running: `process()`
- running -> initialized: (return from process)
- initialized -> instantiated: `deinit()`

tests/test_pipelines/test_asset.py
```diff
@@ -1,29 +1,42 @@
...
+
 
-@pytest.mark.xfail(raises=NotImplementedError)
 def test_runner_scoped():
     """
     say we have an asset that just provides an integer, and the node just increments it.
     we should be able to call Runner.process multiple times, and the asset should
     increment without resetting.
     """
-    raise NotImplementedError
```
sneakers-the-rat: fix docstring - `tube` scoped assets are persistent across process calls"

src/noob/testing/nodes.py
```diff
@@ -9,7 +8,6 @@
 from typing import Annotated as A
 from typing import Any
 
-import xarray as xr
```
sneakers-the-rat: are we no longer using xarray? if so let's drop it as a dependency

src/noob/state.py
```diff
@@ -37,38 +20,45 @@ class State(BaseModel):
...
+    def deinit_assets(self, scope: AssetScope) -> None:
         """
-        asset = [val for key, val in self.assets.items() if key == signal]
-        return None if len(asset) == 0 else asset[-1]
+        run :meth:`.Asset.deinit` for assets that correspond to the given scope.
+        Usually means that :attr:`.Asset.obj` attribute is cleared to `None`.
+        """
+        for asset in self.assets.values():
+            if asset.scope == scope:
+                asset.deinit()
```
sneakers-the-rat: these run four times per process call, so they are perf sensitive and this could get slow with lots of assets, mostly because pydantic attribute access is super slow. making a map from scope -> asset is cheap

src/noob/state.py
```diff
@@ -92,12 +82,36 @@ def collect(self, edges: list[Edge]) -> dict | None:
...
+                        # asset dependency has been satisfied,
+                        # so we don't want to mutate asset anymore
+                        # ideally i'd like to convert whatever this is
+                        # to a copy-on-write object, but it doesn't seem possible.
+                        # https://stackoverflow.com/questions/12359366/python-copy-on-write-behavior
+                        args[edge.target_slot] = deepcopy(asset.obj)
+                    else:
+                        # wrong order of epoch.
+                        # wait until the correct epoch call comes in.
+                        return None
```
sneakers-the-rat: this is sort of backwards I think, and requires extra bookkeeping (the depends_satisfied_epoch). 

so this copies every time the object is accessed downstream of the dependency. this also makes the mutation behavior different pre/post the asset->node dependency.

say we have a chain a->b where a depends on the asset, mutates it by 1, b depends on a and does the same, and then the asset depends on the output of b. 
then say we have c->d such that c depends on the value of b *and also the asset directly*, d depends on the output of c *and the asset directly*. They also mutate the asset by +1.
So a->b->c->d but c and d just request the asset as they would if there was no dependency.

so what we would want is
epoch 0
- a: 0+1=1
- b: 1+1=2 (stored)
- c: 2+1=3
- d: 3+1=4

epoch 1
- a: 2+1=3
- b: 3+1=4
- c: 4+1=5
- d: 5+1=6

what we would get here is
epoch 0
- a: 0+1=1
- b: 1+1=2 (stored)
- c: 2+1=3
- d: **2+1=3**

epoch 1
- a: 2+1=3
- b: 3+1=4
- c: 4+1=5
- d: **4+1=5**

because `d` gets the copied version of the asset. 

so this would make it so the next epoch correctly gets the prior value, but it also changes the behavior of assets within the same epoch, which would be surprising. 

we want to make the copy when the asset is *stored* rather than when it's gotten i think. that's a more predictable behavior and requires less copying.

tests/test_pipelines/test_asset.py
```diff
@@ -0,0 +1,44 @@
...
+    However, a given asset is shared by nodes within a single process call.
+    """
+    tube = Tube.from_specification("testing-process-asset")
+    runner = SynchronousRunner(tube=tube)
+
+    runner.init()
+    for _ in range(5):
+        output = runner.process()
+        assert output["next"] == 1  # renewed each process call
+        assert next(output["counter"]) == 2  # but shared among nodes
```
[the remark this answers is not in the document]
sneakers-the-rat: no just add another node! don't base the test on return!

tests/test_pipelines/test_asset.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.mark.xfail(raises=NotImplementedError)
```
sneakers-the-rat: shouldn't xfail this

tests/test_pipelines/test_asset.py
```diff
@@ -0,0 +1,87 @@
...
+def test_asset_depends_ooo():
+    """
+    when a wrong order epoch begins,
+    the asset should not be injected to the node,
+    hanging the process call.
+    """
+    raise NotImplementedError()
+
+
+@pytest.mark.xfail(raises=NotImplementedError)
```
sneakers-the-rat: also shouldn't xfail this, this is the core of the feature.

tests/test_pipelines/test_asset.py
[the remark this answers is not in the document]
sneakers-the-rat: i.e. the "shared among nodes" part - `return` is special, we want to use a non-special node to test if it's shared among nodes.

src/noob/testing/nodes.py
raymondwjang: let's do it

raymondwjang: fastapi and numpy dropped as well

src/noob/state.py
raymondwjang:
```python
    scope_to_assets: dict[AssetScope, list[Asset]] = Field(default_factory=dict)
```

tests/test_pipelines/test_asset.py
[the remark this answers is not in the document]
raymondwjang: oh i meant after i add another node i need to inspect the tube to see what happened to the assets

src/noob/state.py
sneakers-the-rat: probably a `defaultdict` for convenience's sake

src/noob/state.py
```diff
@@ -92,12 +82,36 @@ def collect(self, edges: list[Edge]) -> dict | None:
...
-                obj = None if asset is None else asset.obj
-                args[edge.target_slot] = obj
+                asset = self.assets[edge.source_signal]
+                if not asset.depends:
+                    args[edge.target_slot] = asset.obj
+                else:
+                    if epoch == asset.depends_satisfied_epoch + 1:
+                        # asset dependency is not satisfied yet
+                        args[edge.target_slot] = asset.obj
+                    elif epoch == asset.depends_satisfied_epoch:
```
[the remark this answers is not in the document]
raymondwjang: there might be cases where we need to store multiple epochs of the asset:

```
epoch 0: a* -> b* -> c -> d -> e
epoch 1: a* -> b* -> c -> d -> e
epoch 2: a* -> b* -> c -> d -> e
```
where `a` depends on an asset and the asset on `b`.

if the nodes `a` and `b` are processed much faster than nodes `c`, `d` and `e`, epoch 2 could technically be entering node `c` before epoch 0 has completed all of nodes `c`, `d`, `e`

this means the "post-depends" asset should be different for all 3 epochs, i think.

src/noob/state.py
sneakers-the-rat: we can ensure that there are maximum 2 copies at a time if we forbid nodes that are in later topological generations than the node that emits the event that the asset depends on from depending on the asset directly, which imo makes sense to do anyway. that would be a check we do when validating the tube specification.

raymondwjang: or did you mean one copy in the event store another copy in the asset

src/noob/runner/sync.py
```diff
@@ -103,9 +104,13 @@ def process(self, **kwargs: Any) -> ReturnNodeType:
                 events = self.store.add_value(node.signals, value, node_id, epoch)
                 if events is None:
                     continue
+                if node_id in self.tube.state.need_copy:
+                    self.tube.state.update(events)
```
sneakers-the-rat: great, much tidier!

src/noob/testing/nodes.py
```diff
@@ -202,3 +191,20 @@ def process(self, left: float, right: float = 1) -> float:
...
+
+
+def jump(generator: count, n: int = 1) -> A[count, Name("skirttt")]:
+    for _ in range(n):
+        next(generator)
+    return generator
+
+
+def rewind(generator: count, n: int = 1) -> A[count, Name("skrittt")]:
+    """Purposely designed to diverge inplace change and output"""
```
sneakers-the-rat:
```suggestion
    """Purposely designed to mutate the input and return a new object"""
```

src/noob/state.py
```diff
@@ -37,38 +20,67 @@ class State(BaseModel):
...
+    See :attr:`.AssetSpecification.depends`
+    """
+    scope_to_assets: dict[AssetScope, list[Asset]] = Field(
+        default_factory=lambda: defaultdict(list)  # type: ignore[arg-type]
+    )
+    """
+    Map from :class:`.AssetScope` to :class:`.Asset` to circumvent
+    querying scope for each asset in :meth:`.State.init_assets` and :meth:`.State.deinit_assets`
+    """
+    need_copy: dict[PythonIdentifier, dict[PythonIdentifier, bool]] = Field(default_factory=dict)
```
sneakers-the-rat:
```suggestion
    need_copy: dict[NodeID, PythonIdentifier] = Field(default_factory=dict)
```

mere presence in this dict indicates copying, right? and the keys are node ids?

src/noob/state.py
```diff
@@ -92,12 +104,24 @@ def collect(self, edges: list[Edge]) -> dict | None:
...
+                    )
 
         return None if not args or all(val is None for val in args.values()) else args
 
+    def update(self, events: list[Event]) -> None:
+        """Update asset if asset depends on a node signal"""
+        for event in events:
+            asset_id = self.dependencies.get(".".join((event["node_id"], event["signal"])))
+            if asset_id is not None:
+                self.assets[asset_id].update(value=event["value"], epoch=event["epoch"])
```
sneakers-the-rat: this being two dicts `needs_update`/`dependencies` is a little clunky to me. could be done with a single `{node_id: {"asset_id": asset_id, "signal": signal}}` mapping.

```python
for event in events:
    if (dep := self.dependencies.get(event['node_id'])) and dep['signal'] == event['signal']:
        self.assets[dep['asset_id']].update(value=event['value'], epoch=event['epoch'])
```

sneakers-the-rat (commented):
cool, getting there. couple things, i'm going to write the validator i mentioned in chat and clean some of these things up then lets do it

src/noob/runner/asyncio.py
sneakers-the-rat: `asyncio` is missing the `update` step, only implemented in sync runner

tests/data/pipelines/asset_depends.yaml
```diff
@@ -0,0 +1,41 @@
...
+  increment:
+    type: noob.testing.fast_forward
+    params:
+      n: 1
+    depends:
+      - generator: assets.counter
+  more_increment:
+    type: noob.testing.fast_forward
+    depends:
+      - generator: assets.counter
```
sneakers-the-rat: this is not the pattern we want to demonstrate or test - the whole deal here is that we want to *pass the asset as events between nodes rather than depending on it directly* so that the dependency structure between nodes controls the order the asset is mutated in. the current state of needing to have *other* dependencies to structure asset mutation is the whole problem and why we need this feature in the first place.

sneakers-the-rat: i'm just going to rewrite this

sneakers-the-rat: https://github.com/miniscope/noob/pull/130/commits/278fa37f74022184f5fb2d82884446ee18f8c917

tests/data/pipelines/asset_depends.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - n: increment.next
+  multiply:
+    type: noob.testing.multiply
+    params:
+      left: 0
```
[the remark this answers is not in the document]
sneakers-the-rat: https://github.com/miniscope/noob/pull/130/commits/278fa37f74022184f5fb2d82884446ee18f8c917

src/noob/state.py
sneakers-the-rat: https://github.com/miniscope/noob/pull/130/commits/278fa37f74022184f5fb2d82884446ee18f8c917

src/noob/state.py
```diff
@@ -37,38 +20,67 @@ class State(BaseModel):
...
-
-    def get(self, signal: str) -> Asset | None:
+        assets = {spec.id: Asset.from_specification(spec) for spec in specs.values()}
+        asset_dependencies = {
+            spec.depends: spec.id for spec in specs.values() if spec.depends is not None
+        }
+        node_dependencies = {".".join((e.source_node, e.source_signal)) for e in edges}
+        need_copy = {}
+        for signal in asset_dependencies:
+            need_copy[signal.split(".")[0]] = {signal.split(".")[1]: signal in node_dependencies}
```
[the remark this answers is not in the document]
sneakers-the-rat: https://github.com/miniscope/noob/pull/130/commits/278fa37f74022184f5fb2d82884446ee18f8c917

src/noob/state.py
sneakers-the-rat: https://github.com/miniscope/noob/pull/130/commits/278fa37f74022184f5fb2d82884446ee18f8c917

src/noob/state.py
```diff
@@ -105,7 +119,11 @@ def collect(self, edges: list[Edge], epoch: int) -> dict | None:
                     "(assets have no generic 'value' signal)"
                 )
                 asset = self.assets[edge.source_signal]
-                if not asset.depends or epoch == asset.stored_at + 1:
+                if (
+                    not asset.depends
+                    or asset.depends.split(".")[0] not in self.dependencies
+                    or epoch == asset.stored_at + 1
```
raymondwjang: i see, so we're not allowing assets to be called directly twice, period (even before it returns from depends node)?

tests/data/pipelines/asset_depends.yaml
```diff
@@ -8,34 +8,27 @@ assets:
...
-      - generator: assets.counter
+      - iterator: a.iterator
   post_depends:
-    type: noob.testing.rewind
-    params:
-      n: -100
+    type: noob.testing.increment
+    depends:
+      - iterator: b.iterator
+  return:
```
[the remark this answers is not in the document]
sneakers-the-rat: what i was saying is that we shouldn't use return nodes *as the nodes that are testing the behavior* - so like if we want to test something is true for nodes in general, we shouldn't use `return` to test that, since return is special. using `return` to return values from the tube is fine.

https://github.com/miniscope/noob/pull/130#discussion_r2700301920

src/noob/state.py
```diff
@@ -105,7 +119,11 @@ def collect(self, edges: list[Edge], epoch: int) -> dict | None:
                     "(assets have no generic 'value' signal)"
                 )
                 asset = self.assets[edge.source_signal]
-                if not asset.depends or epoch == asset.stored_at + 1:
+                if (
+                    not asset.depends
+                    or asset.depends.split(".")[0] not in self.dependencies
```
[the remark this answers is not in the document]
sneakers-the-rat: when there are no other nodes that depend on the signal that the asset depends on - i.e. there are no other ways it could be mutated, so we don't need to copy in order to protect against mutations.

https://github.com/miniscope/noob/pull/130#issuecomment-3762253292

tests/data/special/asset_equal_depends.yaml
```diff
@@ -0,0 +1,35 @@
...
+    params:
+      n: 1
+    depends:
+      - generator: assets.counter
+  multiply:
+    type: noob.testing.multiply
+    params:
+      left: 0
+    depends:
+      - right: more_increment.next
```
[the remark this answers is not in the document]
raymondwjang: or is this not used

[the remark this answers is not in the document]
sneakers-the-rat: ah, it is used in a test that asserts it's invalid, but that test is not specific enough since it just checks for any validation error

tests/test_tube.py
```diff
@@ -66,3 +66,15 @@ def test_cycle_check():
 
     with pytest.raises(ValidationError):
         Tube.from_specification("testing-cycle")
+
+
+@pytest.mark.assets
+@pytest.mark.parametrize("tube", ["testing-invalid-post-depends", "testing-invalid-equal-depends"])
+def test_assets_exhausted_after_storage(tube: str):
```
raymondwjang: oh ok nvm got it

src/noob/state.py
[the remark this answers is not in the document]
raymondwjang: makes sense

PR #136, opened 2026-01-27
refactor zmqrunner to async

### LINKED ISSUE #135: convert zmq runner to full async
currently zmq runner's components are "synchronous with an eventloop in a thread that calls sync callbacks" which turns out to be really inefficient because the process doesn't allocate time in threads very smartly. it also makes lots of fun potentials for thread/eventloop deadlocks. 

so we should refactor the node runners especially to asyncio

sneakers-the-rat:
fix #135 

(just drafting, awkward middle state here, need to pull out all the scheduler awaiting stuff. might be best to just take it out of scheduler altogether)

----
📚 Documentation preview 📚: https://noob--136.org.readthedocs.build/en/136/


sneakers-the-rat:
alright i can’t really replicate this locally, but if you take a look at the codspeed metrics, something like 40% of the time is spent in a set_dealloc function which looks like an internal python method used to deallocate memory used by frames when rendering a traceback? if i remove that from the calculations then the numbers match what i see locally. I don’t see any tracebacks printed and there isn’t a place where they would be getting swallowed without being re-raised that I can see. i’m going to mark those regressions as acknowledged and move on since it is actually faster and this looks like it’s just some artifact from mysterious profiler instrumentation

when measuring locally, this is about twice as fast as the prior implementation, and 4x as fast as https://github.com/miniscope/noob/pull/136/commits/74cf737d544a21781393ae6ef9f5e428f7c81e90 for some reason, not sure why Poller is so slow.

sneakers-the-rat:
doing a merge since raymond is busy!

PR #151, opened 2026-02-20
Assets for zmq runner

### LINKED ISSUE #111: Assets for zmqrunner
Related to:
- #99

sneakers-the-rat:
Fix: #111 

OK General tension here is that on the other runners the assets are literally the same object, so propagating mutation is free. In the ZMQ runner we want to mimic the other runners behavior and behave intuitively as much as is reasonable, but distributed is just fundamentally different than local, it's the whole reason we have events as the driving design pattern, etc.

So the strategy:
- `node` scoped assets are fine, those work as expected
- `process` scoped assets shoudl also work *as expected from the point of view of the graph* - we compute topo layers that all depend on the same asset, we init the asset in the first generation, and then send the asset through the next topo layers. However unlike the local case, mutation is ambiguous within a topo generation: nodes might receive any of the assets from the prior topo generation. this is correct from the graph, but might be counterintuitive. we will have to see what the use cases are for process scoped assets if this proves to be a problem
- `runner` scoped assets could have two desired outcomes: either someone wants to use a runner scoped asset as something that is only initialized once, but it doesn't mutate the asset - in that case we want to keep the same asset through the lifetime of the noderunner. the other is that the asset *is* mutated, and we want to propagate those mutations with as much fidelity as we can. We started assuming the latter case, and assets are propagated similarly to process-scoped assets. We want to overwhelmingly suggest to downstream users that if they want runner-scoped-asset-behavior-with-mutation then they should follow the strategy of exchanging the asset around with events and using the asset dependencies mechanism for persisting state. 

Behind the scenes, we use the `State` object to initiate the assets that the nodes are suppose to init (e.g. because they are the first generation in the asset topo generations), and otherwise treat assets as events, like we do with inputs. This required a little bit of gnarly private method special behavior, the solution to which would be #152 

Along the way fixed a few bugs that came up in these tests specifically:
- serializing a generator would cause pydantic to just iterate through it, and so it would hang infinitely for infinite generators. pickled any iterator that doesn't have a length
- that let us consolidate the serialization which had two near-identical implementations
- the async runner didn't init process-scoped assets
- zmq runner had its epoch ending logic tied up in handling the return node, so it was technically impossible to have a tube without a return value and use the `process` method, which is sort of embarassing. we really need to set up combinatoric testing across tube features * runners * calling patterns, but the zmq runner takes so damn long to bring up.

----
📚 Documentation preview 📚: https://noob--151.org.readthedocs.build/en/151/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,7 @@ jobs:
 
       - name: Coveralls Parallel
         uses: coverallsapp/github-action@v2
+        continue-on-error: true
         with:
           flag-name: run-${{ join(matrix.*, '-') }}
           parallel: true
--- a/docs/api/runner/zmq.md
+++ b/docs/api/runner/zmq.md
@@ -1,7 +1,30 @@
 # `zmq`
 
-```{eval-rst}
+```{eval-rst}              
 .. automodule:: noob.runner.zmq
+```
+
+## Main Runner
+
+```{eval-rst}
+.. automodule:: noob.runner.zmq.runner
+    :members:
+    :undoc-members:
+```
+
+## Command Node
+
+```{eval-rst}
+.. automodule:: noob.runner.zmq.command
+    :members:
+    :undoc-members:
+```
+
+
+## Node Runner
+
+```{eval-rst}
+.. automodule:: noob.runner.zmq.node
     :members:
     :undoc-members:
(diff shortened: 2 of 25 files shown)


sneakers-the-rat:
alright additional line count in the PR is from some verbose tests and docs

PR #156, opened 2026-02-25
so it's come to this, a noob monorepo

sneakers-the-rat:
i want to make a second package `noob-nodes` that will have common noob nodes in it. this will be a namespace package that has different sub-packages for different things like networking and etc. so anyway yeah this is just changing um things to a monorepo.

pretty much the same for now except noob is in a few more layers of nesting. the `packages/noob/src/noob` seems redundant but it's just to make a namespace package. hopefully not too hard to configure IDEs to collapse the indentation and whatnot. 

moved the pdm scripts and the ruff, black, etc. configs up to the top pyproject.toml, but i figure pytest configs might be different per subpackages. so left those down unda.

----
📚 Documentation preview 📚: https://noob--156.org.readthedocs.build/en/156/


sneakers-the-rat:
pretty much all the lines changed are just updating the lockfile btw

PR #167, opened 2026-03-03
Model signals in the topo sorter

### LINKED ISSUE #152: `TopoSorter` dependencies should really be on the `node.signal` pair and not just the node
There are a number of circumstances where we can get events from a given node at different times, most notably assets. we should figure out some way to make dependencies on the topo sorter be `node`->`node.signal` but that has obvious complications re: graph modeling - would we have the node and all its slots in the graph? worth more thought, but the awkwardness is accumulating around not having this.

sneakers-the-rat:
Fix: #152 

This was sort of inevitable: the graph structure of noob is really source_node <- source signal <- target slot <- target, but we're only modeling node <- node in the topo sorter.

This adds explicit source node <- source signal <- target node modeling - excluding target slot modeling for now, that will be needed to handle requiredness better later.

Approach is just to have `(node_id, signal_id)` tuples in the topo sorter such that....

- signal tuples always depend on the source node
- the node is marked as done when any of its signals are emitted
- each of the signals can have separate done/expired conditions

still need to work out the map/gather case, but that's i think just becuase i failed to update how the parent/child epoch creation/expiration works somewhere, i am pretty foggy right now.

Additional changes to make this happen

- Formerly `NoEvent` was emitted with a special `__META__` signal. i don't know why we did that, i assume it was to be able to unambiguously identify `NoEvent`s rather than matching if some node actually emitted a string `NoEvent`. I just added that as a prefix to the string value of NoEvent so it's pretty unlikely to be emitted by accident, and now that it properly emitted with the signal.
- ... other stuff too, will finish writing this when i finish the implementation.

like i said i'm pretty foggy right now, so this is pretty messy, particularly with subepoch handling. 

----
📚 Documentation preview 📚: https://noob--167.org.readthedocs.build/en/167/

### DIFF
--- a/packages/noob/src/noob/exceptions.py
+++ b/packages/noob/src/noob/exceptions.py
@@ -96,12 +96,6 @@ class EpochExistsError(SchedulerError, ValueError):
     """
 
 
-class NotOutYetError(SchedulerError, ValueError):
-    """
-    Node was marked done but wasn't passed out yet!
-    """
-
-
 class NotAddedError(SchedulerError, ValueError):
     """
     Node was marked done but wasn't added!
--- a/packages/noob/src/noob/event.py
+++ b/packages/noob/src/noob/event.py
@@ -84,7 +84,7 @@ class MetaEvent(Event):
 
 
 class MetaSignal(StrEnum):
-    NoEvent = "NoEvent"
+    NoEvent = f"{META_SIGNAL}NoEvent"
 
 
 def _type_discriminator(v: dict | Event | MetaEvent) -> str:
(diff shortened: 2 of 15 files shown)


sneakers-the-rat:
perf regressions are to be expected since we are adding lots more nodes to the graph, but resolving that will be part of a more general scheduler perf optimization once we settle the structure - i think we will be able to eliminate a lot of the weird edge cases by just properly modeling the graph, but still a bit of work to get there.

packages/noob/src/noob/exceptions.py
```diff
@@ -96,12 +96,6 @@ class EpochExistsError(SchedulerError, ValueError):
     """
 
 
-class NotOutYetError(SchedulerError, ValueError):
-    """
-    Node was marked done but wasn't passed out yet!
-    """
```
sneakers-the-rat: it turns out that we suppress this exception whenever it's raised anyway, and we also don't care about this error with the way we use the topo sorter - we routinely want to mark nodes that have not been passed out as done or expired.

PR #168, opened 2026-03-03
length check for sequences when testing if their items are events

### LINKED ISSUE #165: nodes cannot return multiple empty `Sequence`s
temporary issue:

```python
            values = [value] if len(signals) == 1 else value

            new_events = []
            for signal, val in zip(signals, values):
                # Nodes can emit explicit events or collections of events
                if is_event(val):
                    self.add(val)
                    new_events.append(val)
                elif isinstance(val, Sequence) and is_event(val[0]):
```

sneakers-the-rat:
fix: #165 

----
📚 Documentation preview 📚: https://noob--168.org.readthedocs.build/en/168/

### DIFF
--- a/packages/noob/src/noob/store.py
+++ b/packages/noob/src/noob/store.py
@@ -105,7 +105,7 @@ class EventStore:
                 if is_event(val):
                     self.add(val)
                     new_events.append(val)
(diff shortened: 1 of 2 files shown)


raymondwjang (approved):
well that was fast

PR #173, opened 2026-03-06
optional edges, switching tests

### LINKED ISSUE #28: allowing `dict` for `depends`
currently, `TubeSpecification` only allows for lists for `depends`, such as following:

```yaml
  b:
    type: gather
    params:
      n: 5
    depends:
      - value: a.letter
```

and throws a validation error for the `dict` format:
```yaml
  b:
    type: gather
    params:
      n: 5
    depends:
      value: a.letter
```

it feels like allowing a `dict` format might be an expected behavior. thoughts?

sneakers-the-rat:
Fix: #28 

I was trying to write a node that was like

```python
def something() -> tuple[A[str, Name("a")], A[str, Name("b")]]:
    if eh:
        return ah, MetaSignal.NoEvent
    else:
        return MetaSignal.NoEvent, ah
```

and then use it in a tube that basically had two branches like the `branch_switching.yaml` example in the tests, where some downstream node depends on the outcome of both branches, but will be empty when that branch doesn't run. 

that turns out to be impossible because the final stage would never run, since one of the branches would always terminate in a NoEvent.

Previously we had just not included edges in the topo sorter for non-required edges, but that breaks the graph for obvious reasons (the node could run in the same generation as the optional dependency node and just never receive the events). we still have the evaluation of slots -> edges as optional based on their type annotation, it's just currently unused.

So this PR just implemented optional edges when some slot is `type | None = None`.

To make that work, I had to add a bit of ~ graph evaluation ~ such that nodes keep track of their successors and predecessors, and keep a set of successors that are have optional dependencies that they decrement the nqueue of when they get expired. This is to handle the case where the node with the optional slot is >1 hops away from the node that emits a NoEvent.

e.g. say node C here depends on `x` and `y`, but `y` is optional. B emits NoEvent. C should still run, just with `x`

```mermaid
flowchart LR
  A["A(x)"] --> AA["AA(x)"]
  B["B(NoEvent)"] --> BB
  AA --> C
  BB --> C
```

By precomputing the set of nodes we would unblock like this, we avoid needing to propagate NoEvents through the whole tube. 

 This set only includes the immediate layer of nodes with optional deps (rather than all downstream nodes with optional deps) because that's all we can infer from the graph structure, and otherwise we might double-decrement nodes.
 
 The question to me is whether this behavior should be default for `type | None = None` or whether that should be explicit in the dependency annotation, it would be easy to keep the explicit behavior here becasue return and gather both explicitly create their `Slot`s, and we could enable it as the default behavior in the future if we wanted to. i could go either way

----
📚 Documentation preview 📚: https://noob--173.org.readthedocs.build/en/173/


packages/noob/src/noob/toposort.py
```diff
@@ -22,10 +30,22 @@ def __init__(self, node: GraphItem) -> None:
...
-        # long as they're all reflected in the successor's npredecessors attribute.
+        # Immediate successor nodes that run after this node does
         self.successors: set[GraphItem] = set()
 
+        # Immediate predecessor nodes that we depend on to run
+        self.predecessors: set[GraphItem] = set()
+
+        # Optional (immediate) predecessors - we only run once their fate has been decided,
+        # but we can run without them having emitted an event
+        self.optional_predecessors: set[GraphItem] = set()
```
sneakers-the-rat: the reason this is sign flipped (optional rather than required as elsewhere) is because we assume optional deps will be comparatively rarer, we want things to 'fail required,' and also to make the impl clearer to write (checking for stuff that exists rather than stuff that doesn't exist).

sneakers-the-rat: it's awkward but it wasn't just random. i should probably add that to docstring

packages/noob/src/noob/toposort.py
```diff
@@ -360,6 +410,88 @@ def _get_nodeinfo(self, node: GraphItem) -> _NodeInfo:
...
+        for node in expired:
+            self._ready_nodes.discard(node)
+            if node in self._out_nodes:
+                self._out_nodes.discard(node)
+            else:
+                self._npassedout += 1
+        self._done_nodes.update(expired)
+        self._nfinished += len(expired)
+
+    def _update_optionals(
```
sneakers-the-rat: this is so verbose mostly because we don't have a separated "prepare vs. run" phase like the old topo sorter, where if we had a prepare phase then we don't have to go back and correct prior state (the two pass thing) and otherwise handle having incomplete information. i definitely see why the old implementation was the way it was, and also since we aren't even really using dynamic graph editing and just made the map impl work by adding more epochs, we might want to go back to that.

PR #186, opened 2026-03-07
More examples! Nobes image, nobes files, nobes os, and file watcher examples

sneakers-the-rat:
an example of watching a folder, transcoding and image, and putting OCR'd text on a clipboard.

hell a lot of boilerplate, but most of of it is the lockfile.

- `nobes-files` for file manipulation nodes. most notably here the file watcher. Also has some janky ass renaming and moving commands that are sorta contrived, but used for the example. consider these placeholders.
- `nobes-image` for image manipulation, used for OCR and image format conversion
-   `nobes-os` for OS interaction. I was going to have a sweet interactivity demo with system notifications with a button and whatnot, but it turns out mac has locked down the notification center and i can't really test it locally. So this is basically just for the clipboard, but i left in the notifier code in case anyone else wants to pick it up. nobes packages are sort of slop buckets without pretense at being bulletproof.

----
📚 Documentation preview 📚: https://noob--186.org.readthedocs.build/en/186/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# CONTRIBUTING
+
+See [docs/contributing.md](docs/contributing.md)
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,0 +1,23 @@
+# Contributing
+
+```{warning}
+Contributing docs are a work in progress and are not complete!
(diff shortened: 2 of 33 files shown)


sneakers-the-rat:
these are all just examples and raymond still busy! merging!!

Issue #158, opened 2026-02-26
`extends` in tube specifications


sneakers-the-rat:
It should be possible to extend another tube with an `extends` like

```yaml
noob_id: my_tube
extends:
  - another-tube
```

and then the other tube is basically injected into the current tube, with the definitions in the current tube overriding any matching ones in the extended tube.

should be a list because you might want to extend multiple! then the tubes are imported in order, with later entries overriding previous entries

PR #187, opened 2026-03-10
More examples!!!!! web scraping

sneakers-the-rat:
A recursive web scraping example!!! will follow up with a version that splits out the example here into two, where the image scraping does `extends` (#158 ) to add image scraping.

Things fixed/changed in noob in order to make this work:
- properly run async runner in the CLI - there is a decent amount of code duplication here, but i can't really tell how to avoid that, since we're doing a runner * async/sync * input format * output format product of run conditions
- make it possible (default) to not show the progress bar, since that is sort of annoying when you don't want it
- add `flatten` to gather to turn lists of lists into lists
- handle when a collection is length 0 in map by emitting noevents
- limit the max number of outstanding 
- initial implementation of iter for async runner (but see #194)
- limit the max number of tasks in the async runner because for web scraping we would just queue thousands of pages to be scraped and them all of them would hit a timeout
- add async contextmanager `__aenter__` and `__aexit__`
- allow asset params to also be lists (for positional params)
- handle `inputs` for asset params
- allow some minimal destructuring for list of list params by making the input kwargs substitution 
- allow assets to have arbitrary types, like how did we do numpy assets before?
- make config dir into config dirs, because that just sorta makes sense to do
- fix dependency declaration in the store: before we would just not have an entry in the `dependencies` dict if there wasn't another node that depended on an asset that was to be preserved for another epoch. that was done because it was the mechanism by which we avoided deepcopying assets that didn't need to be (i.e. they couldn't be mutated because they had no downstream nodes that received them), but that's a silly way to do that - the dependencies dict should have the dependencies in it. so just keep track of which nodes don't need to be deepcopied. i made it a public set because i can imagine a time when one might want to declare some asset as one that shouldn't be deepcopied.
- prevent NoEvent from being stored as the asset, whoopsie.
- detect async generators in the iscoroutine check function. we generally use that for "is this an async thing," and so this might need further refinement, but it fits with how it's used now.

----
📚 Documentation preview 📚: https://noob--187.org.readthedocs.build/en/187/

### DIFF
--- a/packages/noob/src/noob/input.py
+++ b/packages/noob/src/noob/input.py
@@ -3,7 +3,7 @@ import warnings
 from collections import ChainMap, defaultdict
 from enum import StrEnum
 from functools import cached_property
-from typing import Any, ClassVar, TypeVar, overload
+from typing import Any, ClassVar, overload
 
 from pydantic import BaseModel, ConfigDict, Field, PrivateAttr
 
@@ -11,8 +11,6 @@ from noob.exceptions import ExtraInputWarning, InputMissingError
 from noob.node.base import Edge
 from noob.types import AbsoluteIdentifier, PythonIdentifier
 from noob.yaml import id_optional_json_schema
-
-_TParams = TypeVar("_TParams", bound=list | dict)
 
 
 class InputScope(StrEnum):


raymondwjang (approved):
quicker than i thought! lgtm

PR #191, opened 2026-03-11
json schema - forbid extras in specs

sneakers-the-rat:
following #190 

whoopsie - had this written but not committed. forbid extras in spec objects so that we are alerted when we add extra invalid things!

----
📚 Documentation preview 📚: https://noob--191.org.readthedocs.build/en/191/


sneakers-the-rat:
the performance regression is fake news. how could this cause perf regression

PR #199, opened 2026-03-14
Spiking neural net example

sneakers-the-rat:
here ya go dan, this makes probably the least computational sense of all the demos but here ya go - you can indeed do a spiking neural net where all the neurons run in different processes and spend 99% of the time in ser/des shuttling events back and forth lmao. cycles are actually no problem when you don't care about event alignment that much.

```shell
python -m noob run examples/snn-fully-connected.yaml --runner zmq-freerun
```

----
📚 Documentation preview 📚: https://noob--199.org.readthedocs.build/en/199/

### DIFF
--- a/packages/noob/src/noob/runner/__init__.py
+++ b/packages/noob/src/noob/runner/__init__.py
@@ -1,6 +1,9 @@
 # ruff: noqa I001 - import order meaningful to avoid cycles
+import warnings
+from importlib.metadata import entry_points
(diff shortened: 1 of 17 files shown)


packages/noob/src/noob/runner/__init__.py
sneakers-the-rat: this we'll actually merge - providing runners is a sort of obvious thing to be able to do

PR #202, opened 2026-03-15
`is` comparison to NoEvent

### LINKED ISSUE #192: Can't use `==` equality operator when detecting `NoEvent`
when a value is a numpy array, we can't do equality to check if it is NoEvent!

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

SO we just need to use `is` rather than `==`

will add tests to this effect

sneakers-the-rat:
Fix: #192 
Fix: #201 

----
📚 Documentation preview 📚: https://noob--202.org.readthedocs.build/en/202/

### DIFF
--- a/packages/noob/tests/fixtures/events.py
+++ b/packages/noob/tests/fixtures/events.py
@@ -1,0 +1,24 @@
+from datetime import UTC, datetime
+from typing import Any
+
+import pytest
+
+from noob.event import Event
+from noob.types import Epoch
+
+
+class NonEquivalent:
+    def __eq__(self, other: Any) -> bool:
+        raise ValueError("Cant equals me!")
+
+
+@pytest.fixture(scope="session")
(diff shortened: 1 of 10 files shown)


sneakers-the-rat:
since this is so tiny i'm going to make a couple of readme fixes in this pr ............

packages/noob/tests/fixtures/events.py
```diff
@@ -0,0 +1,24 @@
...
+
+import pytest
+
+from noob.event import Event
+from noob.types import Epoch
+
+
+class NonEquivalent:
+    def __eq__(self, other: Any) -> bool:
+        raise ValueError("Cant equals me!")
```
raymondwjang: equal me if you can ahh

raymondwjang (approved):
lgtm

PR #207, opened 2026-05-01
Live tube visualization

sneakers-the-rat:
a new cli command `noob view my-tube` that displays your tube as you write it!

https://github.com/user-attachments/assets/a7d31e29-06ba-4d6d-b450-046e9d460236

this is the "certified dumbass way to do it" where i am polling with a websocket from inline JS rather than building that into the react component, that'll come next. this is the minimal implementation that just makes it exist, no tests yet becoz not even sure what of this would need to be tested, i guess that the websocket stream works, but i'm gonna hold off on that because we don't have any js tests and that's sorta the real thing we need and that will require a bit of infra. i can port over some of the testing stuff from sciop here because it actually required a bit of doing to get that off the ground.

so what's changed here
- the package build process now builds the JS stuff and bundles it into the source dist. the wheel is built from the sdist so it ends up in there too. i am a little not thrilled at the size of the bundles it ends up producing, but that is sort of 'how it goes' with bundling js. we want to be able to use this with just a pip install. eventually we might want to take advantage of our monorepo lifestyle and make this into a separate package.
- cleaned up the vite configs so all the common stuff is in a base config
- added vite configs for building into the package
- moved the logic of finding the config file into a separate class method because that need comes up a lot and is also something we need for mio
- added `gui` optional dependency group to hold these deps, the gui module fails to load if they are absent
- and then ya the cli command

give it a whirl, to get it going you need to run the js build command which is specified in the js package `npm run watch:package`

----
📚 Documentation preview 📚: https://noob--207.org.readthedocs.build/en/207/


raymondwjang:
Starting to read through, rest assured it will be done by eod tomorrow!

sneakers-the-rat:
Oops sorry for leaving CI a mess

docs/conf.py
```diff
@@ -73,7 +73,7 @@
 
 # autodoc-pydantic
 autodoc_pydantic_model_show_json_error_strategy = "coerce"
-autodoc_pydantic_model_show_json = True
+autodoc_pydantic_model_show_json = False
```
sneakers-the-rat, line 76: it looks like `autodoc-pydantic` broken something upstream where it no longer handles generic types correctly, had to do this in mio too. package might be unmaintained

raymondwjang:
ah crap i reviewed the wrong pr lol

raymondwjang:
will get to this within a couple days sorry

sneakers-the-rat:
discussed offline a bit, raymond has a new job now!

PR #209, opened 2026-05-06
Viz nested tubes

sneakers-the-rat:
Extends: #207 - this goes after that. coupled because i wanted to yno have the things reload while i was working and also there isn't a good way to render the tubes at all otherwise.

Allows visualizing nested tubes!

- Add a pydantic validation context option to load the recursive tube spec
- construct a reactflow version of the tube that puts the recursive tube in a grouping s.t. the tube creates a bordered region with its inputs on the left and returns on the right, and then has its own little nodesies inside of it
- ALSO have to construct an ELK-specific format where the children are re-nested but slightly differently to how we do it.so RELUCTANTLY we are forced to have three forms of the tube in play at once. this should probably get cleaned up into classes but the functional form works for now
- then we lay that shit out and convert it back to the reactflow version

prety much it, other things:

- split up css into element-specific files, this still gets bundled and minified
- moved accessory functions in the flow.tsx file into `tube.tsx` - component files should probably just have components
- add JS tests
- only run nobes tests when we change nobes, it was too noisy, idk what we want to have there.
- correctly order handles on nodes to minimize edge crossings
- some better sizing estimates to give to elk (will refine this further when i get to aesthetic changes)

that gets us to here. as you can see, much to be desired - trying to get all the basic elements working before working on styling things.

Screenshot 2026-05-05 at 9 58 21 PM

----
📚 Documentation preview 📚: https://noob--209.org.readthedocs.build/en/209/


js/src/flow.tsx
```diff
@@ -50,78 +59,234 @@ export function NoobFlow(props: NoobFlowProps) {
...
+
+      if (node.type === "return") {
+        // similarly, return values are on the container
+        targetNode = prefix;
+        targetHandle = `${prefix}.${slot}`;
+      } else {
+        targetNode = `${prefix}.${node.id}`;
+        targetHandle = `${prefix}.${node.id}.${slot}`;
+      }
+    }
```
sneakers-the-rat: couldn't think of a more elegant way to do this, but i'm sure there is.

basically we need special handling of inputs and return types. 

So say we have some node 'b' that is a tube node, and it has its own nodes 'a', 'b', 'c', (the double use of 'b' is intentional). We prefix those like `'b.a', 'b.b', 'b.c'` so that they don't get confused with any nodes a b c in the outer tube (including the parent node itself), so their signals and slots are like `b.a.value`. except to the outer tube, it wants to depend on e.g. `b.value` to get the return value from the tube node, so we have to un-nest references to returns and inputs here like this.

js/src/flow.tsx
```diff
@@ -50,78 +59,234 @@ export function NoobFlow(props: NoobFlowProps) {
...
+        id: id,
+        label: label,
+        key: id,
+      };
+    })
+    .filter((e, index, self) => self.map((x) => x.id).indexOf(e.id) === index);
+  return { sourceHandles, targetHandles };
+}
+
+function isTubeNode(node: NoobNode): node is TubeNode {
```
sneakers-the-rat: handy little typescript dynamic type checking feature

js/src/useLayoutNodes.tsx
```diff
@@ -9,75 +9,109 @@ import { type Edge, useNodesInitialized, useReactFlow } from "@xyflow/react";
...
+// https://eclipse.dev/elk/reference/options/org-eclipse-elk-nodeSize-options.html
 const layoutOptions = {
   "elk.algorithm": "layered",
   "elk.direction": "RIGHT",
-  "elk.layered.spacing.edgeNodeBetweenLayers": "40",
-  "elk.spacing.nodeNode": "40",
-  "elk.layered.nodePlacement.strategy": "SIMPLE",
+  "elk.layered.spacing.edgeNodeBetweenLayers": "20",
+  "elk.spacing.nodeNode": "50",
+  "elk.edgeRouting": "SPLINES",
```
sneakers-the-rat, line 23: it turns out this is important to set in elk, even though we don't use its spline rendering, so that it knows where to place the nodes if one were to do so. eventually we'll want to feed the elk splines into reactflow, but not for this PR.

js/src/useLayoutNodes.tsx
```diff
@@ -97,3 +130,12 @@ export default function useLayoutNodes() {
 
   return null;
 }
+
+interface PropertiedElkNode extends OElkNode {
```
sneakers-the-rat, line 167: hate the naming here, need to extend the types from the package "original elk node" becasue they do also accept properties, the types in the upstream package are just incomplete.

packages/noob/src/noob/node/spec.py
```diff
@@ -95,7 +95,7 @@ class NodeSpecification(BaseModel):
 
     Subclasses should override this with a default.
     """
-    id: PythonIdentifier = Field(..., exclude=True)
+    id: PythonIdentifier
```
sneakers-the-rat, line 98: need these when we're working programmatically, gonna see if this causes problems for regular tube serialization (it shouldn't, just makes them a little more verbose by default)

packages/noob/src/noob/node/spec.py
```diff
@@ -123,7 +123,7 @@ class NodeSpecification(BaseModel):
     description: str | None = None
     """An optional description of the node"""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="forbid", serialize_by_alias=True)
```
sneakers-the-rat, line 126: so that `type_` becomes `type`

packages/noob/src/noob/tube.py
```diff
@@ -41,6 +41,13 @@ class TubeSpecification(ConfigYAMLMixin):
     Those require importing and introspecting the specified node classes,
     which should only happen when we try and instantiate the tube -
     this class is just a carrier for the yaml spec.
+
+    Instance creation can be customized with parameters passed to the model's
+    validation context
+    (see: https://pydantic.dev/docs/validation/latest/concepts/validators/#validation-context ).
+
+    * ``recursive: True`` - replace the tube ID in a :class:`.TubeNode` with the full spec
```
sneakers-the-rat, line 49: all the changes here and below in `yaml.py` are to enable this - just gives us control over how tubes are loaded, so we can e.g. recursively load tube definitions when we want to but not have to do that always.

sneakers-the-rat:
most of the bulk here comes from moving and splitting files btw

sneakers-the-rat:
pressin on with the gui push

PR #211, opened 2026-05-08
make slot and signal accessors class methods

sneakers-the-rat:
we need this for the GUI, and for spec reflection generally - currently one needs to instantiate the nodes in order to get their signals and slots, because some nodes have dynamically computed signals and slots that are dependent on their specification, and others like the wrap nodes need to locate the appropriate callable to treat as the process method. however this is likely to be pretty fragile for cases where we want to be operating on the metadata level re: a tube - some downstream node could do a lot of work on initialization, and we don't want to have to do it when all we want is to tell what some node emits and receives.

so this just makes them classmethods that accept a spec - by default the spec isn't needed, but subclasses may require it if they do.

we still need a convenience accessor to go from spec -> node class -> signals/slots, that logic for deciding what node class to use is still wrapped up in the tube creation process, but i'll do that in another PR.

----
📚 Documentation preview 📚: https://noob--211.org.readthedocs.build/en/211/

### DIFF
--- a/packages/noob/src/noob/node/base.py
+++ b/packages/noob/src/noob/node/base.py
@@ -52,7 +52,9 @@ class Slot(BaseModel):
         slots = {}
         sig = inspect.signature(func)
 
-        for name, param in sig.parameters.items():
+        for i, (name, param) in enumerate(sig.parameters.items()):
+            if i == 0 and name == "self":
+                continue
             slots[name] = Slot(
                 name=name,
                 annotation=param.annotation,
@@ -74,6 +76,8 @@ class Signal(BaseModel):
         return_annotation = inspect.signature(func).return_annotation
         for name, type_ in cls._collect_signal_names(return_annotation):
             signals.append(Signal(name=name, type_=type_))
+        if not signals:
+            signals = [Signal(name="value", type_=Any)]
         return signals
 
     @classmethod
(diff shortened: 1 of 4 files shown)


packages/noob/src/noob/node/base.py
```diff
@@ -294,21 +298,44 @@ def from_specification(
...
+        Public class method for computing signals from a node class and a specification.
+
+        Exposed as a class method for reflection purposes -
+        we don't want to have to instantiate a node to tell what would come out of it,
+        but signals can be dynamically computed by the node depending on its spec.
+
+        If signals can't be computed (because they are not annotated, etc.),
+        returns a generic "value" signal that refers to whatever the node produces.
+
+        Base implementation reads annotations from the process method and does not use the spec.
```
raymondwjang, line 322: do we raise an error if spec is not the same as the process method annotations?

sneakers-the-rat: That probably wouldn't be here, that would be more of a spec validation thing. 

Like by default the spec isn't used at all and we just do inspection of the function signature. The spec is passed for the wrap nodes (and in case downstream nodes want to use it) where the spec is needed to compute the signals and slots (e.g. the wrap nodes need to know what they are wrapping)

sneakers-the-rat: We could do that with the validation context keywords, and it probably should be done when going from tube spec -> tube when we have all the relevant things imported, but it would be a pretty expensive validation to do whenever we load a tube spec.

sneakers-the-rat: If you take a look at the nodeinfo PR #213 then that might be a mechanism for that, since it adds a lazy coupling between the spec and the node info

PR #214, opened 2026-05-12
Use Nodeinfo when visualizing to draw slots and signals

sneakers-the-rat:
the backlog groweth bigger

Builds on: #211, #212, #213

Previously we just took the spec's word for it and just drew whatever edges existed in it, but that was strictly wrong - nodes have other signals and slots than what the spec might use, and also the spec could be wrong and just making shit up.

so we just forward the node info along with the node and that simplifies things a lot.

also
- more unification of signal and slot: rename `type_` to `annotation`
- mode the edge classes (signal, slot, edge) to their own module for import reasons

----
📚 Documentation preview 📚: https://noob--214.org.readthedocs.build/en/214/

### DIFF
--- a/js/src/tube.tsx
+++ b/js/src/tube.tsx
@@ -20,7 +20,30 @@ import type { Edge } from "@xyflow/react";
  */
 export function tubeToFlow(tube: TubeSpecification): [Edge[], NodeUnion[]] {
   const edges = getEdges(tube.nodes);
-  const nodes = getNodes(tube.nodes, edges);
+  let nodes = getNodes(tube.nodes);
+  // TODO: Dedicated representation of inputs
+  if (tube.input) {
+    nodes = [
+      ...nodes,
+      {
+        id: "input",
+        position: { x: 0, y: 0 },
+        type: "elk",
+        data: {
+          label: "input",
+          targetHandles: [],
+          sourceHandles: Object.values(tube.input).map((i) => {
+            return {
+              id: `input.signals.${i.id}`,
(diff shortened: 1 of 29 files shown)


js/src/tube.tsx
```diff
@@ -20,7 +20,30 @@ import type { Edge } from "@xyflow/react";
...
+              id: `input.signals.${i.id}`,
+              label: i.id,
+              key: `input.signals.${i.id}`,
+              required: true,
+            };
+          }),
+        },
+      },
+    ];
+  }
```
sneakers-the-rat, line 46: this is sort of a shame, but we're currently treating inputs as just being a node, which is wrong. this is at least a step away from that

js/src/tube.tsx
```diff
@@ -109,12 +132,8 @@ function getNodeEdges(node: NoobNode, prefix?: string): Edge[] {
 }
 
 // Get all nodes from a tube spec
-function getNodes(nodes: Record<string, NoobNode>, edges: Edge[]): NodeUnion[] {
-  const has_input = edges.some((e) => e.source === "input");
-  if (has_input) {
-    nodes = { ...nodes, input: { id: "input", type: "input" } };
-  }
-  return Object.values(nodes).flatMap((node) => getNode(node, edges));
+function getNodes(nodes: Record<string, NoobNode>): NodeUnion[] {
```
sneakers-the-rat, line 135: removing dependencies is good - getting the nodes should only depend on the nodes

PR #224, opened 2026-05-20
Perf: avoiding epoch.__eq__

sneakers-the-rat:
doing some simple profiling and perf improvements, `Epoch.__eq__` was taking an astonishingly long time, 6-7% of runtime for the kitchen sink with the async runner in 1000 process calls.

Screenshot 2026-05-19 at 7 32 32 PM

turns out that can totally be eliminated in normal runs by just reordering things s.t. we don't need to call `==` when we are given an epoch. comments inline

----
📚 Documentation preview 📚: https://noob--224.org.readthedocs.build/en/224/

### DIFF
--- a/packages/noob/src/noob/scheduler.py
+++ b/packages/noob/src/noob/scheduler.py
@@ -6,7 +6,6 @@ from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from functools import cached_property
-from itertools import count
 from typing import Self
 from uuid import uuid4
 
@@ -33,7 +32,7 @@ class Scheduler:
     source_nodes: list[NodeID] = field(default_factory=list)
     _logger: logging.Logger = field(default_factory=lambda: init_logger("noob.scheduler"))
 
-    _clock: count = field(default_factory=count)
+    _last_epoch: int = -1
     _epochs: dict[Epoch, TopoSorter] = field(default_factory=dict)
     _subepochs: dict[Epoch, set[Epoch]] = field(default_factory=lambda: defaultdict(set))
     _epoch_log: deque[int] = field(default_factory=lambda: deque(maxlen=100))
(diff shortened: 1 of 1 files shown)


packages/noob/src/noob/scheduler.py
```diff
@@ -95,19 +94,20 @@ def add_epoch(self, epoch: int | Epoch | None = None) -> Epoch:
...
+            elif this_epoch in self._epoch_log:
+                raise EpochCompletedError(f"Epoch {this_epoch} has already been completed!")
+
             # ensure that the next iteration of the clock will return the next number
             # if we create epochs out of order
-            self._clock = count(
-                max([this_epoch[0].epoch, *[ep[0].epoch for ep in self._epochs], *self._epoch_log])
-                + 1
-            )
+            self._last_epoch = max(self._last_epoch, this_epoch[0].epoch)
```
sneakers-the-rat, line 107: avoiding needing to iterate through all epochs to find the max - we only need to consider the last epoch we've created and have an integer rather than an iterator.

packages/noob/src/noob/scheduler.py
```diff
@@ -95,19 +94,20 @@ def add_epoch(self, epoch: int | Epoch | None = None) -> Epoch:
...
         else:
-            this_epoch = Epoch(next(self._clock))
-
-        if this_epoch in self._epochs:
-            raise EpochExistsError(f"Epoch {this_epoch} is already scheduled")
-        elif this_epoch in self._epoch_log:
-            raise EpochCompletedError(f"Epoch {this_epoch} has already been completed!")
```
sneakers-the-rat, line 110: avoid needing to use eq when searching through the epoch log (deque, epochs is a dict and key lookup is O(1)) - we only need to check this when we are passed an epoch explicitly. if we are just making 'the next one' then we can trust the internal counter (someone can come along and purposely fuck this up that that's on them lol)

packages/noob/src/noob/scheduler.py
```diff
@@ -408,14 +413,14 @@ def epoch_completed(self, epoch: Epoch) -> bool:
         return previously_completed or active_completed
 
     def end_epoch(self, epoch: Epoch | int | None = None) -> MetaEvent | None:
-        if epoch is None or epoch == -1:
+        if isinstance(epoch, Epoch):
+            ep = epoch
```
sneakers-the-rat, line 418: simple reordering

PR #230, opened 2026-06-14
perf: replace epoch_log deque with ordered dict for O(1) lookup

### LINKED ISSUE #229: Use sets for `epoch_log`
we use `epoch_log` to track if epochs have already been run while being able to gc the epochs themselves. this is necessary because we want to error if trying to run the same epoch again after it's already been done, that protects us and signals deeper errors in the scheduler - for the async and sync runners it's mostly not necessary, but for the zmq runner it is. We can't just keep an int here because epochs can happen out of order.

we don't care about the order, the only reason we do is that we want to keep it from growing infinitely, but honestly i think that might be overoptimizing: epochs are tiny:

using the recursive `getsizeof` recipe linked from the docs - https://code.activestate.com/recipes/577504-compute-memory-footprint-of-an-object-and-its-cont/

```python
>>> total_size(Epoch(1))
193
>>> total_size(Epoch(1_000_000_000_000))
197
>>> total_size(Epoch((('root', 1), ('areallylongname', 1), ('anevenlongernameholycrapitkeepsgoing', 1))))
454
```

so at a million epochs you'd hit ~180MB or so. we only really need to keep a log of the epochs that could be coming in out of order, so that might be in some worst case the last ~1000, so we could just do a periodic trim, once every thousand epochs, keep the last 100 epochs, or something, idk.

anyway, iterating through a list is O(n) and checking set membership is O(1) so we should switch that up somehow.

vaishnavidesai09:
Closes #229

## What
Replaces `_epoch_log: deque` with `_epoch_log: dict` (insertion-ordered).

## Why
- `in` checks on a deque are O(n); dict membership is O(1)
- `min(epoch_log)` was O(n); `next(iter(epoch_log))` gets the oldest key in O(1)
- dicts are insertion-ordered, so we can trim oldest epochs cheaply by slicing keys

## How
- `_epoch_log[ep[0].epoch] = True` replaces `.append(ep[0].epoch)`
- Membership checks now compare against plain `int` keys consistently
- Periodic trim every `_epoch_log_trim_interval` completions (default 1000),
  keeping the `_epoch_log_keep` most recent (default 100)
- Trim happens after insert so the new epoch is always retained
- Out-of-order epochs are safe — trim removes by insertion order, not by value

## Tests added
- `test_epoch_log_is_set_like` — completed epochs are in the log, future ones aren't
- `test_epoch_log_trim_keeps_recent_epochs` — trim evicts old epochs, keeps most recent
- `test_epoch_log_out_of_order_trim` — late-arriving high-numbered epoch is never evicted

## Test results
All pre-existing tests pass. One pre-existing failure (`test_noevent_ends_epoch`) 
is an intentional placeholder on main and unrelated to this PR.

----
📚 Documentation preview 📚: https://noob--230.org.readthedocs.build/en/230/

### DIFF
--- a/packages/noob/src/noob/scheduler.py
+++ b/packages/noob/src/noob/scheduler.py
@@ -230,7 +230,7 @@ class Scheduler:
         # which marks them as "out" in the TopoSorter
 
         # if we've already run this, the node is ready - don't create another epoch
-        if epoch in self._epoch_log:
+        if epoch and epoch[0].epoch in self._epoch_log:
             return True
 
         graphs = (
@@ -243,7 +243,7 @@ class Scheduler:
 
     def node_is_done(self, node: NodeID, epoch: Epoch) -> bool:
         """Node is expired or done in specified epoch"""
-        if epoch in self._epoch_log:
+        if epoch[0].epoch in self._epoch_log:
             return True
 
         if self._subepochs[epoch]:
@@ -346,8 +346,7 @@ class Scheduler:
         Args:
             with_signals (bool): When marking this node as done, also mark all its signals as done.
         """
-        epoch_int = epoch[0].epoch if isinstance(epoch, Epoch) else epoch
-        if epoch_int in self._epoch_log:
+        if epoch[0].epoch in self._epoch_log:
             self._logger.debug(
                 "Marking node %s as done in epoch %s, " "but epoch was already completed. ignoring",
                 node_id,
@@ -410,7 +409,7 @@ class Scheduler:
         previously_completed = (
             len(self._epoch_log) > 0
             and epoch not in self._epochs
-            and (epoch_int in self._epoch_log or epoch_int < next(iter(self._epoch_log)))
+            and (epoch_int in self._epoch_log or epoch_int < min(self._epoch_log))
         )
         active_completed = epoch in self._epochs and not any(
             self._epochs[ep].is_active() for ep in [epoch, *self._subepochs[epoch]]
--- a/packages/noob/src/noob/scheduler.py
+++ b/packages/noob/src/noob/scheduler.py
@@ -472,7 +472,7 @@ class Scheduler:
         Remove epoch records, restarting the scheduler
         """
         self._epochs = {}
-        self._epoch_log = {}
+        self._epoch_log = set()
 
     def _init_graph(self, epoch: Epoch | None = None) -> TopoSorter:
         """
--- a/packages/noob/src/noob/scheduler.py
+++ b/packages/noob/src/noob/scheduler.py
@@ -35,8 +35,8 @@ class Scheduler:
     _last_epoch: int = -1
     _epochs: dict[Epoch, TopoSorter] = field(default_factory=dict)
     _subepochs: dict[Epoch, set[Epoch]] = field(default_factory=lambda: defaultdict(set))
-    _epoch_log: dict[int, None] = field(default_factory=dict)
-    _epoch_log_trim_interval: int = field(default=1000)
+    _epoch_log: set[int] = field(default_factory=set)
+    _epoch_log_trim_interval: int = field(default=200)
     _epoch_log_keep: int = field(default=100)
     _subgraphs: dict[NodeID, tuple[dict[str, NodeSpecification], list[Edge]]] = field(
         default_factory=dict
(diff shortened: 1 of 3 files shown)


vaishnavidesai09:
Hi @sneakers-the-rat, I've opened a PR for this!

Went with the ordered dict approach — keys are plain ints (ep[0].epoch), 
values are True. This gives O(1) membership and lets us trim from the front 
by insertion order cheaply, so out-of-order high-numbered epochs are never 
accidentally evicted. Trim is periodic (every 1000 completions, keep last 100) 
rather than on every insert to avoid unnecessary iteration.

Added 3 tests covering membership, trim behavior, and out-of-order safety. 
Happy to adjust the trim interval/keep defaults or sketch the set approach 
for comparison if useful!

sneakers-the-rat:
great! looks pretty much like what we want. i'll just have to do a few things to integrate it with some other outstanding PRs (that i thought were merged already) and then get it in :)

packages/noob/src/noob/scheduler.py
```diff
@@ -403,10 +405,11 @@ def epoch_completed(self, epoch: Epoch) -> bool:
         """
         Check if the epoch has been completed.
         """
+        epoch_int = epoch[0].epoch
         previously_completed = (
             len(self._epoch_log) > 0
             and epoch not in self._epochs
-            and (epoch in self._epoch_log or epoch < min(self._epoch_log))
+            and (epoch_int in self._epoch_log or epoch_int < min(self._epoch_log))
```
sneakers-the-rat, line 412: needs to be min to handle out of order problems (see added test `test_epoch_completed_out_of_order`).

the inequality works with `epoch` rather than `epoch_int` because `Epoch` implements inequality with ints, but doing the inequality on the int is more direct.

packages/noob/src/noob/scheduler.py
```diff
@@ -426,7 +429,11 @@ def end_epoch(self, epoch: Epoch | int | None = None) -> MetaEvent | None:
             raise TypeError("Can only end an epoch with an integer or Epoch")
         self._logger.debug("Ending epoch %s", ep)
         if len(ep) == 1:
-            self._epoch_log.append(ep[0].epoch)
+            self._epoch_log[ep[0].epoch] = None
+            if len(self._epoch_log) >= self._epoch_log_trim_interval:
+                self._epoch_log = {
+                    k: None for k in sorted(self._epoch_log)[-self._epoch_log_keep :]
+                }
```
sneakers-the-rat: maybe counterintuitively, recreating the object tends to be faster than deleting from it in a loop

```python
def _time_del():
    original = {k: None for k in range(1000)}
    for to_delete in sorted(original)[:900]:
        del original[to_delete]

def _time_recreate():
    original = {k: None for k in range(1000)}
    new = {k:None for k in sorted(original)[-100:]}

>>>  %timeit _time_del()
82.6 μs ± 257 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

>>> %timeit _time_recreate()
63.4 μs ± 2 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

sneakers-the-rat:
Confirming that the perf regression from CodSpeed is an artifact in the noise floor. when profiling the three approaches on the pure scheduler benchmark we get this:

(with max-time of 10s, and with keep_size=100 and clear_interval=200):
| version | best time | run time | iters | iter/s |
| --- | --- | --- | --- | --- | 
| `main` | 2.8us | 9.33s | 166,345 | 17,829 |
| `dict` | 2.64us | 9.82s | 171,540 | 17,468 |
| `set` | 2.46us | 10.00s | 189,164 | 18,916 |

the set and the dict are pretty interchangeable codewise, just one line difference with `add` vs. an assignment. weirdly, the dict version seems to be marginally slower than the deque, i assume that all comes from reconstructing the epoch log with only the most recent epochs - this is a behavioral improvement over `main` because main might drop off out of order epochs sooner, but recreating the dict is like 20x as costly as a single iteration of the whole epoch ( https://github.com/miniscope/noob/pull/230#discussion_r3416601712 ).

the set approach seems to be cheaper to recreate, for a modest but not-insubstantial 6% performance improvement for the scheduler - every % counts!

sneakers-the-rat (approved):
Thanks for the PR! this also looks like it fixes a few places where we were checking `Epoch` against an int too.

PR #231, opened 2026-06-25
Consolidate scheduling logic in scheduler

### LINKED ISSUE #144: Consolidate epoch tracking to scheduler
The zmq runner has two mechanisms for keeping track of current epoch, one in the ZMQRunner and another in the Noderunner. the other runners also just keep a counter to iterate through epochs. 

This makes #140 happen, and also makes the `NodeRunner.await_inputs` method hopelessly complex. 

The scheduler should be the source of information for what nodes can run and when, and it should encapsulate the implementation of statefulness so that it's uniform across runners. Then we just return to having a single `get_ready()` method that we can iterate over and trust, rather than having to do a bunch of finnicky stuff around it.

We should make sure we propagate statefulness from the node instantiation to the spec (since it can be determined from node defaults as well as explicit setting in the spec) so that the scheduler has it, and then the scheduler should be made aware of it: when a node is stateful, we don't yield it as ready in future epochs until it has been completed in prior epochs.

sneakers-the-rat:
fix: #144 

i've gotta run atm, but briefly, the public API of the scheduler is *way* too big, and runners are expected to manage a ton of scheduling logic.

want to replace most of the publicly exposed methods with two iterators, `iter_epoch` and `iter_ready`, one iterates a single epoch, the other iterates infinitely until no nodes are ready to be run (in this case runners are expected to handle waiting).

just getting started adding between-epoch dependencies to the topo sorter and scheduler, then will go back and remove the scheduling logic from the topo sorter and mark most of the methods private.

ok wrote the pr notes in the changelog:

**Perf**
- Cached {attr}`.Node.signals` and {attr}`.Node.slots` as they are frequently accessed props
- More of the {class}`.TopoSorter`'s methods were adapted to be batched set operations
  rather than iterators.
- {class}`.Epoch` 's `__eq__` method was removed to avoid using it during lookups,
  using hash instead and normal tuple equality.
  Being able to `==` an integer wasn't worth the performance cost.
- {class}`.Epoch`'s creation was optimized to avoid remaking {class}`.EpochSegment`s.
- {class}`.Epoch` gained `__slots__`

**Added**
- `stateful` is propagated from {class}`~.Node` classes back to {class}`~.NodeSpecification` s.
  `stateful` can be specified explicitly in the spec, or computed automatically by the node class.
  So we need to make sure to propagate that with the spec for e.g. the `ZMQRunner`
- {meth}`~.Scheduler.iter_epoch` and {meth}`~.Scheduler.iter_ready` methods
  to iterate over a single epoch's nodes, or any nodes, respectively, as they are ready.
  This will largely replace the previous pattern of repeatedly calling `get_ready` 
  within a `while scheduler.is_active()` loop.
- A special `('meta', 'previous_epoch')` signal is added to {class}`.TopoSorter`s
  to model statefulness and inter-epoch dependencies within the graph directly.
  Nodes that are `stateful` will depend on this signal, 
  which is marked as done when the previous epoch completes.
  This ensures that nodes that must be run with ordered inputs are in all contexts,
  and dramatically simplifies scheduling while also making concurrent runners
  naturally to handle multiple epochs simultaneously with stateless tubes.
- A {attr}`.Scheduler.epoch` property was added to get the current epoch.
- {meth}`.TopoSorter.get_state` convenience method to get a summary of a topo sorter's state
  (for debugging purposes).
  This copies the internal sets to protect them from mutation,
  and so shouldn't be used in any perf-sensitive paths.
  

**Changed**
- Moved scheduling logic out of runners and into the scheduler!
  Runners used to have to internally keep their own counters of the current/active epochs,
  which was a huge mess and a bad division of labor.
  Now the relationship between a runner and the scheduler is much less janky.
- The ZMQRunner was simplified and its terrible state management was streamlined.
  There are still some flaky test failures from threading deadlocks, but we are getting there.
- the {class}`.TubeRunner`'s `_get_ready` method was made an iterator, yielding from the scheduler.
- The {class}`.TopoSorter` no longer returns `NodeSignal`s as "ready".
  Instead it marks them as being "out" along with the node, returning only the node as ready.
  Previously, they could be returned with a warning in the event of a bug or an
  inconsistent scheduling state that marked a node as done without marking its signals as done.
  Signals can't be run, it makes no sense for them to be ready.
  They are always out at the same time that a node is,
  even if they can have different ending conditions
  (i.e. being marked "done" by emitting an event or "expired" by emitting `NoEvent`)
- The {class}`.Scheduler`'s node completion methods all return a list of {class}`.MetaEvent`s
  rather than one or `None`, allowing multiple epochs to be marked as completed at once.
- An epoch is considered completed by {meth}`.Scheduler.epoch_completed` 
  *only* when it is fully complete, rather than when no more progress can be made,
  but a completion event has not been emitted. 
  This clarifies the distinction and use between it and {meth}`.Scheduler.is_active`,
  which returns `False` when an epoch *should* be marked as completed 
  because no more progress can be made, 
  while `epoch_completed` checks if that completion event has been emitted. 

**Fixed**
- The `ZMQRunner` now more correctly locks the state around its scheduler,
  avoiding race conditions between the command node thread and the main thread.
- Events in the parent epoch that are from a node that is *not* the one that induced subepochs
  can mark their counterparts in subepochs as expired.
  This is how it was supposed to work, and how `gather`ing works - 
  To decrease cardinality, a node emits an event in the parent.
- Empty sets are no longer created in the Scheduler's `_subepochs` dict for every epoch,
  even when there are no subepochs.
- Dynamically enabling/disabling a node with the scheduler now works in epochs that have already been scheduled,
  however this means that nodes should only be enabled/disabled in between running epochs,
  or else nodes may be re-ran in those epochs.
- Even stateless tubes run in a "best effort" epoch order -
  when a bunch of epochs are queued up out of order,
  the scheduler ensures that ready nodes are yielded in an epoch-sorted order.

**Removed**
- {class}`.zmq.NodeRunner` 's `await_inputs` and `await_node` methods 
  were removed in favor of the scheduler iterators.

**CI**
- pytest hooks were added to set the debug logging level when re-running an action in debug mode.

----
📚 Documentation preview 📚: https://noob--231.org.readthedocs.build/en/231/


sneakers-the-rat:
lets see what this says

@codspeedbot fix this regression

sneakers-the-rat:
@codspeedbot fix this regression

sneakers-the-rat:
dumbass bot doesn't know how to even do the thing it tells me to tell it to do

sneakers-the-rat:
Alrighty, had time for review, merging now bc need to keep on truckin.

PR #233, opened 2026-07-01
Represent assets in viewer, fix prefix nesting

sneakers-the-rat:
Working on merging some of the mio changes - 

super simple change, just adding a representation for assets in the viewer, and fixing prefixed nesting of ids

----
📚 Documentation preview 📚: https://noob--233.org.readthedocs.build/en/233/


sneakers-the-rat:
perf regression is just random noise, nothing here would touch that.

PR #244, opened 2026-07-06
Allow multiple assets to be updated from same node

### LINKED ISSUE #200: support updating multiple assets from same node
shortsighted on my part...

i need to change how the `self.tube.state.dependencies` is represented, from
`{node_id: (asset_id, signal)}`

to

`{asset_id: (node_id, signal)}`

so that multiple `runner` scoped assets can be updated from the same node

sneakers-the-rat:
Fix: #200 

Previously, it was only possible to have one asset dependency mapping per node. However nodes can emit multiple signals, and the same signal might be used to update multiple assets (not exactly sure when this would be useful, but it's a logical implication of the spec and should work).

Simple enough change, using a nested dict to keep the external lookups quick (O(1) is node relevant to the update, as used in the zmq runner) - alternative considered was using `NodeSignal()` as the key but that requires allocating a tuple on every lookup.

added `dependency_assets` cached prop to get the set of assets with dependencies since the `dependencies` dict is no longer 1:1 count of assets in the value field.

----
📚 Documentation preview 📚: https://noob--244.org.readthedocs.build/en/244/

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -127,6 +127,10 @@
 - [`238`](https://github.com/miniscope/noob/pull/238) - 
   No more validation errors on type annotations on node process functions from the
   {class}`~noob.edge.Signal` class.
+- [`#200`](https://github.com/miniscope/noob/issues/200),
+  [`#244`](https://github.com/miniscope/noob/pull/244) - 
+  Allow multiple assets to be updated from the same node,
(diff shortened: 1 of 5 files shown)


sneakers-the-rat:
perf regression is fake news, genuinely hate that thing

PR #245, opened 2026-07-06
Propagate NoEvents from nested tubes

### LINKED ISSUE #162: `NoEvent`, `return` and `TubeNode`
When a node within a tubenode emits a `NoEvent`,
this should deactivate all subsequent nodes in the same epoch, but the nodes that depend on the tubenode itself (and subsequently downstream) are not deactivated.

for example, the following combination of recursive tubes:
```yaml
noob_id: parent

nodes:
  motion:
    type: cala.nodes.prep.Anchor
    depends:
    - frame: lines.frame
  downsample:
    type: tube
    params:
      tube: cala-downsample
    depends:
    - frame: motion.frame
  glow:
    type: cala.nodes.prep.GlowRemover
    depends:
    - frame: downsample.value
```
```yaml
noob_id: child

input:
  frame:
    scope: process
    type: cala.assets.Frame

nodes:
  gather:
    type: gather
    params:
      n: 4
    depends:
    - input.frame
  downsample:
    type: cala.nodes.prep.downsample
    depends:
    - frames: gather.value
  return:
    type: return
    depends: downsample.frame
```
causes `glow` to raise: "missing argument: 'frame'" error.

When a `TubeNode`'s `return` node doesn't run _because of_ `NoEvent`, it should propagate `NoEvent` to the parent tube.

### LINKED ISSUE #239: Return `NoEvent`s from tube nodes
runner `process` methods return `None` when no return value was given, and the tube nodes just return this as an actual `None`.

need to
- detect when a tube has returns
- if none is returned, return NoEvent

Or, generally, return `NoEvent` from runner process methods to disambiguate a scalar `None` that's an actual value given to `return`

sneakers-the-rat:
Fix: #162 
Fix: #239 

Interpret a `None` return value from a nested tube as a `NoEvent` in a parent.

This is the gentlest way to do this, without changing the default behavior of tubes to return `NoEvent` when no events made it to the return node (which we could do later, for consistency's sake) - this keeps the output of a runner falsy when no events are emitted.

two line change, the rest is docs and tests. could have reused existing tubes for this, but wanted to have a specific test for this, so it's more verbose than it could have been

edit: it turns out we needed a bit more, to actually propagate the signals from the return node -> tube node. Normally return nodes have no signals, because their values can't be returned within the tube (they are a "level shift" that brings events up to the calling scope). However when they are used in nested tubes, the return values become signals available in the parent tube. So by correctly inferring those and using those for NoEvent, we are basically fixing unimplemented but implied features of the tube node without adding special cases.

----
📚 Documentation preview 📚: https://noob--245.org.readthedocs.build/en/245/


packages/noob/src/noob/node/tube.py
```diff
@@ -104,3 +134,31 @@ def get_slots(cls, spec: NodeSpecification | None = None) -> dict[str, Slot]:
...
+        from noob.tube import TubeSpecification
+
+        if not spec.params or "tube" not in spec.params:
+            raise ValueError("Tube node specifications must have a `tube` in their params")
+        tube_spec = TubeSpecification.from_any(spec.params["tube"])
+
+        signals = {}
+        return_nodes = [n for n in tube_spec.nodes.values() if n.type_ == "return"]
+        if not return_nodes:
+            return {"value": Signal(name="value", annotation=Any)}
```
sneakers-the-rat, line 153: maybe this should just be empty, no point in just always emitting NoEvent as if there was something to subscribe to

sneakers-the-rat: actually nah, we do want to emit the NoEvent in case other nodes want to just use it for ordering with an optional dep

Issue #247, opened 2026-07-07
Deprecate implicit runner `init`s


sneakers-the-rat:
Currently the initialization state can be implicit, calling process/run/iter without explicitly initializing the tube.

We should insist this is explicit:

as a contextmanager (preferred)
```python
runner = SomeRunner(tube)
with runner:
    runner.process()
    # or
    for result in runner.iter():
        # ...
    # or
    runner.run()
```

or manually:

```python
runner = SomeRunner(tube)
runner.init()
runner.process()
runner.deinit()
```

Issue #206, opened 2026-04-28
"soft stop"


sneakers-the-rat:
we need to be able to stop, but finish processing whatever events we've already received - drain the queue. source nodes just stop emitting, and then add flag in iteration to quit when scheduler is empty

Issue #147, opened 2026-02-20
Handle `StopIteration` in generator nodes


sneakers-the-rat:
When generator nodes raise `StopIteration` errors we just crash. we should return from `iter` and `run` and then raise informative error when trying to run again.

PR #248, opened 2026-07-07
Handle generator exhaustion

sneakers-the-rat:
A draft of fixing #147 - but we should not merge this, in working on this i realized that our stopping logic in general needs some work, related to #206 as well.

So this was part of the "see if fable can actually do stuff" test run, the initial commit is its attempt and called it good. clearly that is inadequate. 

mine adds handling exhaustion as an event but is similarly hacky - the things we need are
- #247 - better control over initialization condition, we need a general way to shut it down and exist from whatever run is happening.
- move the event handling into the scheduler, so the scheduler is the thing that says "can't run anymore bub" - so rather than a million different handlers in the different runners, if the scheduler comes across an exhaustion event, it yields the remaining nodes that can run, and then future calls to the ready iterators raise an exception that says "can't run no more"
- more coherent hook handling in general across the runners - we could do with a bit of cleanup, they are sort of diverging from one another and the hook methods aren't actually usable because of that divergence. each of the runners does have a different design, but still it shouldn't be this hard to handle an event.

opening this up as a draft, but we should fix the other things first, and i'll need to think about how we want to handle this in the scheduler.

----
📚 Documentation preview 📚: https://noob--248.org.readthedocs.build/en/248/

### DIFF
--- a/packages/noob/src/noob/__init__.py
+++ b/packages/noob/src/noob/__init__.py
@@ -2,14 +2,14 @@
 
 from noob.config import config as cfg
 from noob.logging import init_logger
+from noob.types import Epoch, Name
+from noob.event import Event, NoEventable, MetaEvent, MetaSignal
 from noob.node import process_method, Node, NodeSpecification
 from noob.asset import Asset, AssetSpecification
 from noob.state import State
 from noob.input import InputCollection, InputScope, InputSpecification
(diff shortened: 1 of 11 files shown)


sneakers-the-rat:
Superceded: #256

